PR #3, opened 2022-02-15
docs: update README.MD

vuvth:
# Proposals

- Improve Unirep setup documents on README.md 
- Fix verify the reputation proof instruction on README.md


vivianjeng:
I am thinking because the proof and the epoch key might change and proof even can be used once
so we don't have to export them to the environment

how about export the proof and publicSignals to a file
and we import them like `-p proof.json`

but we can keep `USER_PRIVATE_KEY` , `ATTESTER_PRIVATE_KEY` , `UNIREP_CONTRACT_ADDRESS` , `USER_IDENTITY` , `IDENTITY_COMMITMENT`  exporting to the environment

good suggestion 🙂

vuvth:
Because the we created `Unirep/docs`, so i close this PR and create new one.

PR #5, opened 2022-03-12
feat:use TypeChain to generate typescript smart contract class.

vuvth:
## Proposal
- Use TypeChain to generate TS smart contract class. 
## TODO

Please review my PR! Thanks <3


vivianjeng:
We can export `Unirep__factory` in `@unirep/contracts`
Then use this factory to replace `deployUnirep` and `getUnirepContract`.

vivianjeng:
`proofIndexes` in [unirep/cli/userStateTransition.ts](https://github.com/vuvth/Unirep/blob/feat/use-typechain/packages/unirep/cli/userStateTransition.ts#L115) should be `BigNumberish` type.

PR #16, opened 2022-04-04
Fix test scripts

vivianjeng:
- use `.test.ts` file name in test scripts
- use different config to check test scripts
- `maxUsers = 10` in test scripts to prevent from timeout
- update readme


packages/circuits/test/utils.ts
```diff
@@ -648,7 +648,6 @@ const genReputationCircuitInput = async (
         epoch: epoch,
```
vuvth, line 648: Why we push utils.ts in test?  Do we reuse method in this file.
if we reuse I think it more sense if we move it to `src` or rename it to index.ts.

packages/contracts/src/index.ts
```diff
@@ -419,27 +419,23 @@ const deployUnirep = async (
...
+    const _maxReputationBudget = _settings?.maxReputationBudget
+        ? _settings.maxReputationBudget
+        : MAX_REPUTATION_BUDGET
+    const _epochLength = _settings?.epochLength
+        ? _settings.epochLength
+        : EPOCH_LENGTH
+    const _attestingFee = _settings?.attestingFee
+        ? _settings.attestingFee
+        : ATTESTTING_FEE
+    _attestingFee
```
vuvth, line 438:
```suggestion
```

packages/contracts/src/index.ts
```diff
@@ -419,27 +419,23 @@ const deployUnirep = async (
...
-        _attestingFee = _settings.attestingFee
-    } else {
-        _maxUsers = MAX_USERS
-        _maxAttesters = MAX_ATTESTERS
-        _numEpochKeyNoncePerEpoch = NUM_EPOCH_KEY_NONCE_PER_EPOCH
-        _maxReputationBudget = MAX_REPUTATION_BUDGET
-        _epochLength = EPOCH_LENGTH
-        _attestingFee = ATTESTTING_FEE
-    }
+    const _maxUsers = _settings?.maxUsers ? _settings.maxUsers : MAX_USERS
```
vuvth, line 422:
```suggestion
    const _maxUsers = _settings?.maxUsers ?? MAX_USERS
```

vuvth: Let do same thing with another config value.

packages/circuits/package.json
```diff
@@ -11,7 +11,7 @@
         "buildProveReputationSnark": "./scripts/buildProveReputationSnark.sh",
         "buildProveUserSignUpSnark": "./scripts/buildProveUserSignUpSnark.sh",
         "copyCircom": "rm -rf ./dist/zksnarkBuild && cp -rf ./zksnarkBuild ./dist/zksnarkBuild",
-        "test": "mocha -r ts-node/register test/*.ts --exit",
+        "test": "mocha -r ts-node/register test/*.test.ts --exit",
```
vuvth, line 14: Why we don't use `hardhat test *.test.ts`?

packages/contracts/test/Airdrop.test.ts
```diff
@@ -14,14 +14,6 @@ import {
...
-    ATTESTTING_FEE,
-    EPOCH_LENGTH,
-    MAX_ATTESTERS,
-    MAX_REPUTATION_BUDGET,
-    MAX_USERS,
-    NUM_EPOCH_KEY_NONCE_PER_EPOCH,
-} from '@unirep/config'
 
 describe('Airdrop', function () {
     this.timeout(100000)
```
vuvth, line 19: Set `this.timeout(0)` if you want disable timeout.

vuvth (commented):
Thank! I left few comments. Please check it

packages/circuits/test/utils.ts
vivianjeng: because it is just used for tests and it will not be used in @unirep/contracts and @unirep/core anymore.

packages/circuits/package.json
vivianjeng: hardhat is not a dependency of circuits

packages/contracts/test/Airdrop.test.ts
vivianjeng: I think it still needs a timeout here

vuvth (approved):

PR #17, opened 2022-04-05
Replace `js-sha256` in `ZkIdentity`

vivianjeng:
- `js-sha256` will cause error in frontend and it will be updated in [@zk-kit/identity](https://github.com/appliedzkp/zk-kit/pull/25)

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -4,21 +4,21 @@
 
 import { ethers } from 'ethers'
 
-export const ATTESTTING_FEE = ethers.utils.parseEther('0.1')
+export const ATTESTTING_FEE = ethers.utils.parseEther('0')
 
-export const NUM_EPOCH_KEY_NONCE_PER_EPOCH = 4
+export const NUM_EPOCH_KEY_NONCE_PER_EPOCH = 3
 
-export const NUM_ATTESTATIONS_PER_PROOF = 4
+export const NUM_ATTESTATIONS_PER_PROOF = 5
 
 export const EPOCH_LENGTH = 30 // 30 seconds
 
-export const GLOBAL_STATE_TREE_DEPTH = 8
+export const GLOBAL_STATE_TREE_DEPTH = 9
 
-export const USER_STATE_TREE_DEPTH = 8
+export const USER_STATE_TREE_DEPTH = 9
 
-export const EPOCH_TREE_DEPTH = 64
+export const EPOCH_TREE_DEPTH = 32
 
-export const MAX_REPUTATION_BUDGET = 5
+export const MAX_REPUTATION_BUDGET = 10
 
 export const MAX_USERS = 2 ** GLOBAL_STATE_TREE_DEPTH - 1
 
--- a/packages/contracts/contracts/Unirep.sol
+++ b/packages/contracts/contracts/Unirep.sol
@@ -140,7 +140,7 @@ contract Unirep is IUnirep, zkSNARKHelper , Hasher, VerifySignature {
     
     function verifyAttesterFee() private view {
         require(
-            msg.value == attestingFee,
+            msg.value >= attestingFee,
             "Unirep: no attesting fee or incorrect amount"
         );    
     }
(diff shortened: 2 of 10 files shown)


packages/contracts/contracts/Unirep.sol
```diff
@@ -140,7 +140,7 @@ contract Unirep is IUnirep, zkSNARKHelper , Hasher, VerifySignature {
     
     function verifyAttesterFee() private view {
         require(
-            msg.value == attestingFee,
+            msg.value >= attestingFee,
```
vuvth, line 143: Thanks!

packages/config/src/index.ts
```diff
@@ -4,21 +4,21 @@
 
 import { ethers } from 'ethers'
 
-export const ATTESTTING_FEE = ethers.utils.parseEther('0.1')
+export const ATTESTTING_FEE = ethers.utils.parseEther('0')
```
vuvth, line 7: Why you change default attesting fee to zero?

packages/contracts/test/Airdrop.test.ts
```diff
@@ -22,18 +22,18 @@ describe('Airdrop', function () {
...
         accounts = await hardhatEthers.getSigners()
 
+        const _treeDepths = getTreeDepthsForTesting()
         const _settings = {
             attestingFee,
         }
-        const _treeDepths = getTreeDepthsForTesting()
 
         unirepContract = await deployUnirep(
             <ethers.Wallet>accounts[0],
```
vuvth, line 39: I see this logic reuse in all files, let create function for it. 
And store it in test/setup.ts.
You can update in current PR or create other PR for it as you refer.

packages/contracts/test/Airdrop.test.ts
```diff
@@ -299,9 +300,7 @@ describe('Airdrop', function () {
             expect(wrongAttestingFee).not.equal(attestingFee_)
```
vuvth, line 300: I don't think we need this assert because it always false. Please check it again for sure. Thanks

packages/core/cli/deploy.ts
```diff
@@ -66,14 +66,11 @@ const deploy = async (args: any) => {
     const _maxReputationBudget = MAX_REPUTATION_BUDGET
```
vuvth, line 66: Why you don't use MAX_REPUTATION_BUDGET directly? Because we don't reassign `_maxReputationBudget` anymore.
Uhm, maybe I wrote this line.
 Should we remove it?

packages/crypto/src/identity.ts
```diff
@@ -1,4 +1,8 @@
-import { ZkIdentity, Strategy } from '@zk-kit/identity'
+import { BigNumber } from '@ethersproject/bignumber'
+import { randomBytes } from '@ethersproject/random'
+import _sha256 from 'crypto-js/sha256'
+import { poseidon } from 'circomlibjs'
+
 import { SnarkBigInt } from './crypto'
 
 type EddsaPrivateKey = Buffer
```
vuvth, line 8: I think we should store all types in one files(types.ts). Let do it in other PR!

vuvth (commented):
Looks good! Left few comments! Thanks for your PR!

packages/config/src/index.ts
vivianjeng: Because in alpha test we pay for the attestingFee and it costs too much and not necessary now.
and it should be worked with `attestingFee=0`

vivianjeng: Because in alpha test we pay for the attestingFee and it costs too much and not necessary now.
and it should be worked with `attestingFee=0`

packages/core/cli/deploy.ts
vivianjeng: yeah there is no need here

vivianjeng: yeah there is no need here

vuvth: Sure

vuvth (approved):

Issue #25, opened 2022-05-12
JSON changes


vimwitch:
Pull the JSON serialization changes from the frontend into the monorepo

The unirep class is overridden [here](https://github.com/Unirep/Unirep-Social-frontend/tree/alpha/src/overrides).

PR #30, opened 2022-05-13
fix: merge frontend overridden function:  `fromJSON`, `toJSON`

vivianjeng:
Fix https://github.com/Unirep/Unirep/issues/25
1. add `fromJSON` and update `toJSON` in `UnirepState` and `UserState`
2. rename `genUserStateFromContract` to `genUserState` (since `genUserStateFromParams` is removed)
   ```typescript
      // genUserState usage
      const userState1 = await genUserState(provider, address, id) // from startBlock
      const userState2 = await genUserState(provider, address, id, userState.toJSON()) // restore userState and update
      const userState3 = UserState.fromJSON(id, userState.toJSON()) // restored userState, no update
      // genUnirepState usage
      const unirepState1 = await genUnirepState(provider, address) // from startBlock
      const unirepState2 = await genUnirepState(provider, address, unirepState.toJSON()) // restore unirepState and update
      const unirepState3 = Unirep.fromJSON(unirepState.toJSON()) // restored unirepState, no update
   ```
3. Use `JSON.stringify(attestation)` instead of `attestation.toJSON()`. Remove duplicated `Attestation`, `IAttestation` as in @unirep/contracts
4. Fix function description
5. Update test scripts

### DIFF
--- a/packages/core/test/utils.ts
+++ b/packages/core/test/utils.ts
@@ -129,7 +129,7 @@ const verifyProcessAttestationsProof = async (
 
 const getReputationRecords = (id: ZkIdentity, unirepState: UnirepState) => {
     const currentEpoch = unirepState.currentEpoch
-    let reputaitonRecord = {}
+    const reputaitonRecord = {}
     for (let i = 0; i < currentEpoch; i++) {
         for (
             let j = 0;


vimwitch:
In the future could you separate refactors and tasks into different commits? In 5960e1d there are a lot of changes; I want to review the `fromJSON`/`toJSON` stuff carefully but didn't expect the refactor in that commit 😅

packages/core/src/UserState.ts
```diff
@@ -9,26 +10,29 @@ import {
...
+import { IAttestation, Attestation } from '@unirep/contracts'
+import { NUM_ATTESTATIONS_PER_PROOF } from '@unirep/config'
+
 import {
     defaultUserStateLeaf,
     genEpochKey,
     genNewSMT,
     genEpochKeyNullifier,
     genReputationNullifier,
+    stringifyAttestation,
```
vuvoth, line 22: `IAttestation` defined in @unirep/contracts. I think this more sense if we move `stringifyAttestation` to `@unirep/contracts`. How you think about it?

packages/core/src/UserState.ts
```diff
@@ -9,26 +10,29 @@ import {
...
     attesterId: BigInt
     reputation: Reputation
 }
 
 interface IReputation {
-    posRep: BigInt
-    negRep: BigInt
-    graffiti: BigInt
-    signUp: BigInt
+    posRep: BigNumber
```
vuvoth, line 32: Shall we about creating a new module(file) for `Reputation`? I think the project structure will be a little bit easier to read.

packages/core/test/utils.ts
```diff
@@ -307,6 +320,100 @@ const genProveSignUpCircuitInput = async (
     return stringifyBigInts(circuitInputs)
 }
 
+const compareObjectElements = (obj1: any, obj2: any) => {
+    let same = true
+    for (const key of Object.keys(obj1)) {
+        if (typeof obj1[key] === 'function') continue
+        if (key === 'unirepState') {
```
vuvoth, line 327: Could we use `assert.deepEqual` ?

vuvoth (commented):
Quickly review! Thank for your PR!

vivianjeng:
@vimwitch would you mind review the final `core/src/UnirepState.ts` and `core/src/UserState.ts` files? Or I can create another PR. Sorry for the inconvenience 😅

packages/core/src/UserState.ts
vivianjeng: I think only `core` requires stringify function so that I put them in `core`.
Or it can be extended from `@unirep/contracts`

vimwitch:
LGTM, i think there are some differences in the args for the JSON stuff, but we can fix it when we integrate it into the frontend.

packages/core/src/Reputation.ts
```diff
@@ -0,0 +1,77 @@
+import assert from 'assert'
+import { BigNumber, BigNumberish } from 'ethers'
+import { hash5, hashOne } from '@unirep/crypto'
+
+import { IReputation } from './interfaces'
+
+export default class Reputation implements IReputation {
+    public posRep: BigNumber
```
vuvoth, line 8: If we don't modify value,  why we don't use readonly.

packages/core/test/utils.ts
```diff
@@ -121,24 +131,28 @@ const getReputationRecords = (id: ZkIdentity, unirepState: UnirepState) => {
     const currentEpoch = unirepState.currentEpoch
     let reputaitonRecord = {}
```
vuvoth, line 132:
```suggestion
    const reputaitonRecord = {}
```

vuvoth (approved):
lgtm!

packages/core/src/interfaces.ts
```diff
@@ -0,0 +1,58 @@
...
+    update(
+        _posRep: BigNumber,
+        _negRep: BigNumber,
+        _graffiti: BigNumber,
+        _signUp: BigNumber
+    ): IReputation
+    addGraffitiPreImage(_graffitiPreImage: BigNumber): void
+}
+
+export interface IUserState {
```
vuvoth, line 49: I think we should move IAttester to this files.

packages/core/src/interfaces.ts
```diff
@@ -0,0 +1,58 @@
...
+    latestEpochKeyToAttestationsMap: { [key: string]: string[] }
+    nullifiers: string[]
+}
+
+export interface IUserStateLeaf {
+    attesterId: BigInt
+    reputation: IReputation
+}
+
+export interface IReputation {
```
vuvoth, line 33: I also see we use same Interface in circuits/test/utils.
https://github.com/Unirep/Unirep/blob/fix/json-serialization/packages/circuits/test/utils.ts#L34

packages/core/src/Reputation.ts
vivianjeng: it could be updated

packages/core/src/interfaces.ts
vivianjeng: `IAttestation` is different from `IReputation`

PR #31, opened 2022-05-13
fix: use matching versions instead of explicit workspace versions

vimwitch:
We need to publish the packages to npm before we can use the monorepo url as a git dependency. [Yarn workspaces will automatically link packages if the specific version is the same as the version in the local package](https://yarnpkg.com/features/workspaces#what-does-it-mean-to-be-a-workspace). This means any external package using the git repo as a package dep will get a version number for other packages in the monorepo.

We need to publish versions of each package to npm so that module resolution won't fail when e.g. `contracts` depends on `crypto`.


vimwitch:
Closing in favor of #32

Issue #19, opened 2022-04-07
@unirep/contracts and @unirep/core should be independent of circuits


vivianjeng:
- Now, contracts and core takes circuits from specific version
- Contracts package should take any key, either from uploaded file/ local built file. It should be independent of circuits version
- Unirep package should build and deploy verifiers from uploaded keys/ local built keys. It should be independent of circuits, contracts version

PR #34, opened 2022-05-19
Refactor: no fix circuit path

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

Issue Number: [#19](https://github.com/Unirep/Unirep/issues/19)

## What is the new behavior?

- **`@unirep/circuits`**
    - Add `UnirepCircuit` object which takes `zkFilePath` as input
    - Integrate circuit utils in the `UnirepCircuit` object
    - Example
        ```typescript
        import circuit from '@unirep/circuits'
        // genProof
        const { proof, publicSignals } = await circuit.genProof(zkFilePath, circuitName, circuitInputs)
        // verifyProof
        const isValid = await circuit.verifyProof(zkFilePath, circuitName, proof, publicSignals)
        ```
    - zkFile directory includes circuit config, users can get circuit config from `zkFilesPath`
- **`@unirep/contracts`**
    - Add `UnirepContract` object which takes `artifactsPath` as input
    - Build contract verifiers before deploy (`yarn contracts build`). The `zkFilesPath` is set in `contracts/scripts/buildVerifiers.ts`
    - Example
      ```typescript
        import contract from '@unirep/contracts'
        // deploy
        const deployedContract = await contract.deploy(artifactsPath, deployer, config)
        // connect unirep contract
        const connectedContract = contract.get(deployedContract.address, provider)
        ```
- **`@unirep/core`**
    - Add `UnirepProtocol` object with `zkFilePath` as input
        - Integrate circuit related functions in the protocol object
        - Example
         ```typescript
         import { UnirepProtocol } from '@unirep/core'
         // initialize protocol object
         const protocol = new UnirepProtocol(zkFilePath)
         // take generating an epoch key for example
         const epochKey = protocol.genEpochKey(idNullifier, epoch, nonce)
         ```
    - `UnirepState` is inherited from `UnirepProtocol`. 
        - It can use the `UnirepProtocol` functions related to circuit config
        - `UnirepState` should be configured from `zkFilePath` as well
        - Example
        ```typescript
         import { UnirepState } from '@unirep/core'
         // initialize unirep state object
         const state = new UnirepState(zkFilePath)
         ```
    - `UserState` is inherited from `UnirepState`. 
        - It can use the `UnirepProtocol` functions related to circuit config
        - It can use the `UnirepState` functions (e.g. `genGST()`, `currentEpoch`, `genEpochTree()`...)
        - `UserState` should be configured from `zkFilePath` as well
        - Example
        ```typescript
         import { UserState } from '@unirep/core'
         // initialize user state object
         const user = new UserState(zkFilePath, id)
         ```
    - `genUnirepState` and `genUserState` takes `UnirepProtocol` object to configured from circuit config
        - Example
         ```typescript
          import { genUnirepState, genUserState, UnirepProtocol } from '@unirep/core'
          // initialize protocol object
          const protocol = new UnirepProtocol(zkFilesPath)
          // generate unirep state object from contract
         const unirepState = await genUnirepState(
             protocol, 
             provider, 
             contractAddress
         )
         // generate user state object from contract
         const userState = await genUserState(
             protocol,
             provider,
             contractAddress,
             id
         )
         ```

- [x] Yes
- [ ] No


packages/contracts/src/contracts/ProcessAttestationsVerifier.ts
```diff
@@ -0,0 +1,118 @@
+/* Autogenerated file. Do not edit manually. */
```
vimwitch, line 1: We should gitignore typechain files. They should be automatically generated at build time instead.

packages/contracts/src/abis/Unirep.ts
```diff
@@ -0,0 +1,1595 @@
+export const UnirepABI = [
```
vimwitch, line 1: I also think we should gitignore contract ABIs. We can auto-generate this file when we generate typechain files instead.

packages/circuits/src/UnirepCircuit.ts
```diff
@@ -0,0 +1,157 @@
...
+     * @param circuitName The name of the circuit
+     * @param inputs The input of the proof
+     * @returns The proof and the public signals of the snark proof
+     */
+    public static genProof = async (
+        zkFilesPath: string,
+        circuitName: CircuitName,
+        inputs: any
+    ) => {
+        const circuitWasmPath = path.join(zkFilesPath, `${circuitName}.wasm`)
```
vimwitch, line 90: The more i think about this I don't think it's worth doing this in a utility class. Instead we can use the standard snarkjs tooling at the application level.

For example, I changed the unirep social backend to not use the circuits package at all and instead use [this](https://github.com/Unirep/Unirep-Social-backend/blob/anondb-sync/src/daemons/Prover.ts). I think it's good to encourage applications to become more comfortable interacting with zk files.

packages/circuits/src/UnirepCircuit.ts
```diff
@@ -0,0 +1,157 @@
+import path from 'path'
```
vimwitch, line 1: This module will have trouble with browser support.

packages/contracts/src/ReputationProof.ts
```diff
@@ -0,0 +1,83 @@
...
+            _publicSignals[this.config.maxReputationBudget + 7]
+        this.proof = formattedProof
+        this.publicSignals = _publicSignals
+    }
+
+    public verify = (): Promise<boolean> => {
+        const proof_ = UnirepCircuit.formatProofForSnarkjsVerification(
+            this.proof.map((n) => n.toString())
+        )
+        return UnirepCircuit.verifyProof(
```
vimwitch, line 55: I'm not a fan of this structure. The dependency structure here looks like this:

`app` -> `@unirep/contracts` -> `@unirep/circuits` -> `snarkjs`

I think it's reasonable to to a more minimalist approach like this:

`app` -> `snarkjs`

It seems like wrapping `snarkjs` with a helper function (like this) makes things more simple, but it actually makes things harder because developers have to learn a wrapper function name/args for every project (e.g. unirep/interep/zkopru/semaphore). Then if there's a bug the developer has to figure out if it's the app, the unirep wrapper logic, or snarkjs. It's much easier to just get a proof and give it to snarkjs IMO.

packages/contracts/src/EpochKeyProof.ts
```diff
@@ -0,0 +1,49 @@
+import { BigNumberish, utils } from 'ethers'
+import { SnarkProof } from '@unirep/crypto'
+import UnirepCircuit, { CircuitName } from '@unirep/circuits'
+
+import { UnirepTypes } from './contracts/IUnirep'
+import { rmFuncSigHash } from './utils'
+import { UnirepABI } from './abis/Unirep'
+
+export class EpochKeyProof implements UnirepTypes.EpochKeyProofStruct {
+    public globalStateTree: BigNumberish
```
vimwitch, line 10: I think there's some value in taking a proof and unwrapping the property names like here, but I don't know if it's worth maintaining this code.

Instead could we improve the documentation for the proofs saying what each index in the proof is?

e.g.

```
EpochKeyProof:
[
  GST,
  epoch,
  epochKey,
]
```

This would let us get rid of these classes instead of having to maintain more code/docs.

packages/contracts/src/UnirepContract.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Get Unirep contract from a specific address
+     * @param address The address of the Unirep contract
+     * @param signerOrProvider A given signer or provider
+     * @param artifactsPath If artifacts path is given, it loads abi from the given artifacts path.
+     * Or it loads Unirep abi from exported package
+     * @returns The Unirep contract object
+     */
+    public static get = (
```
vimwitch, line 121: I think we're falling into the trap of classes that don't really abstract here. Why would I learn this:

```
import UnirepContract from '@unirep/contracts`
import { UNIREP_ADDRESS } from './constants'

const signer = {create signer}
const contract = UnirepContract.get(address, signer)
```

When this does the same thing:

```
import { ethers } from 'ethers'
import { UnirepABI } from '@unirep/contracts'
import { UNIREP_ADDRESS } from './constants'

const signer = {create signer}
const unirepContract = new ethers.Contract(UNIREP_ADDRESS, UnirepABI, signer)
```

Even though we get some type support with typescript it's still sometimes hard to tell what a variable is. If I get an object of type `UnirepContract` it's hard to tell that it's just an ethers contract. The code in the second example is very familiar; I immediately know what I can and can't do with the `unirepContract` object. This is another case of letting developers use shared tooling (ethers) so they can learn the code more quickly.

packages/contracts/src/UnirepContract.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * Deploy Unirep contract and verifiers with compiled artifacts
+     * @param artifactsPath The path to the artifacts (e.g. ../build/artifacts/)
+     * @param deployer The deployer of the contract
+     * @param config The config of the contracts and its circuits
+     * @returns The Unirep contract object
+     */
+    public static deploy = async (
```
vimwitch, line 33: I think most app developers don't need this function. Really it should be a script intended to be run on a dev machine.

vimwitch (changes requested):
I haven't looked at the test changes, just the export changes.

packages/circuits/src/UnirepCircuit.ts
vivianjeng: yeah that makes sense

vivianjeng:
@vimwitch Let me summarize what you think and what we need now
and I propose some thoughts to see if we can do that

It seems like we don't really need `@unirep/circuits` and `@unirep/contracts` only has to export smart contracts
- `@unirep/circuits`
  because what users need `genProof` and `verifyProof` can be replace by `snarkjs`
  so our `@unirep/circuits` can just do the building circuits and testing stuff
  and we will upload `.vkey.json`, `.zkey`, `.wasm` to cloud
  (as for `formatProof` it can be done in `@unirep/core` or user's app)

- `@unirep/contracts`
  users also don't need `deploy` and `get` methods
  so if apps want to 
  1. deploy unirep contract: clone the unirep repo and run `yarn contracts deploy`
  2. get unirep contract: `const contract = new ethers.Contract(UNIREP_ADDRESS, UnirepABI, signer)`
  
  and I am thinking about if we export smart contracts to the package
  (so users can call unirep contract like:
  ```solidity
  import { Unirep } from "@unirep/contracts/Unirep.sol";
  ```
  )
  the type of the typechain will be fixed and abi will be fixed
  also their factories and bytecode are removed
  for those users who don't compile contracts (e.g. unirep-social-frontend)
  they can simply refer to the unirep contract's type
  ```typescript
  import { UnirepABI, Unirep } from '@unirep/contracts'
  const unirep = new ethers.Contract(UNIREP_ADDRESS, UnirepABI, provider) as Unirep
  ```

  Or if all contracts, verifiers will be upload to cloud as well, we don't need `@unirep/contracts`
  and users call unirep contract like:
  ```solidity
  import { Unirep } from "../USER/LOCAL/PATH/TO/Unirep.sol";
  ```
  and use the contract like:
  ```typescript
  import { UnirepABI, UnirepFacotry } from "../USER/LOCAL/COMPILED/ABI/AND/TYPECHAIN/PATH"
  const unirep = UnirepFactory.connect(UNIREP_ADDRESS, provider)
  ```
  
  so what we will upload is
  - circuits
    - .vkey.json
    - .zkey
    - .wasm
  - contracts
    - all unirep solidity contracts (including verifiers)
  
  how do you think?

PR #32, opened 2022-05-17
Published packages

vimwitch:
- Fix package versions to allow local linking
- Change all package versions to `1.0.0`
- Remove `@unirep/config` package, use `@unirep/circuits/config` during testing

PR #37, opened 2022-05-22
Crypto backward compat

vimwitch:
Refactor accessor functions in `ZkIdentity` for backward compatibility.

### DIFF
--- a/packages/circuits/test/identityCommitment.test.ts
+++ b/packages/circuits/test/identityCommitment.test.ts
@@ -21,8 +21,8 @@ describe('(Semaphore) identity commitment', function () {
         )
 
         const id: ZkIdentity = new ZkIdentity()
-        const nullifier = id.getNullifier()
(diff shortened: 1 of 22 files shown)


vimwitch:
depends #32

vivianjeng:
I am trying to use the [`@zk-kit/identity`](https://github.com/privacy-scaling-explorations/zk-kit/tree/main/packages/identity)
but since the sha256 package they used is outdated (and npm hasn't updated yet), I still rewrite it in `@unirep/crypto`.
will I ask them to update npm package?

vimwitch:
@vivianjeng Yea i think it's a good idea. It would be nice if they used getters like in this PR. Functions like `getNullifier()` make it seems like some calculation is happening when really it's just a variable being read.

vivianjeng (approved):
Yeah I am convinced
we can keep getters in `@unirep/crypto` and wait zk-kit updated

PR #45, opened 2022-06-09
Update crypto packages

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

- crypto package depends on `circomlibjs` and `maci-crypto` but most of the crypto functions are not used in unirep
- `IncrementalMerkleTree` and `SparseMerkleTree` use different method names

Issue Number: N/A

## What is the new behavior?

- Write crypto functions that are used in Unirep protocol but importing the packages
- Let `0` be the default leaf value in `IncrementalMerkleTree`
- import `poseidon` code from `circomlibjs`
- Use the same method names in `IncrementalMerkleTree` and `SparseMerkleTree`
  (`tree.root`, `tree.createProof()`, `tree.verifyProof()`)
- Write down crypto test scripts and update README.md
- Modify all other packages that depend on `@unirep/circuits`

- [x] Yes
- [ ] No


packages/crypto/src/IncrementalMerkleTree.ts
```diff
@@ -10,7 +16,9 @@ export class IncrementalMerkleTree extends zkIncrementalMerkleTree {
      * @param zeroValue Zero values for zeroes.
      * @param arity The number of children for each node.
      */
-    constructor(depth: number, zeroValue: any, arity?: number) {
```
vimwitch, line 13: This export is a a monkey patch on top of `zk-kit`. Instead of doing this I think it would be better to remove this export entirely and have downstream code depend directly on `zk-kit`.

packages/crypto/src/SparseMerkleTree.ts
```diff
@@ -22,44 +20,38 @@ const newWrappedPoseidonT3Hash = (
...
-        return tree
-    }
+    protected _root: bigint
+    private zeroHashes!: bigint[]
+
+    public readonly numLeaves: bigint
 
-    constructor(protected db, private height: number) {
-        assert(height > 0, 'SMT height needs to be > 0')
+    constructor(protected db, private _height: number, zeroHash: bigint) {
```
vimwitch, line 28: I'm not sure what `db` argument is here. I know that a `keyv` instance is passed here but not much else. I don't like `keyv` because it doesn't compile well for browsers, it has a lot of nodejs specific requirements.

Could we instead use the `zk-kit` `SparseMerkleTree` implementation?

packages/crypto/src/crypto.ts
```diff
@@ -1,18 +1,103 @@
...
-    stringifyBigInts,
-    unstringifyBigInts,
-} from 'maci-crypto'
+import assert from 'assert'
+import crypto from 'crypto'
+const { stringifyBigInts, unstringifyBigInts } = require('ffjavascript').utils
+
+import { poseidon } from './poseidon'
+
+// Copy from maci-crypto@0.9.1
```
vimwitch, line 7: Why are we copying the `maci-crypto` code here? I think re-exporting it to simplify use is okay, but copying it means we don't get any upstream patches or fixes.

packages/crypto/src/crypto.ts
```diff
@@ -1,18 +1,103 @@
...
+ * @param right The seconde element to be hashed
+ */
+const hashLeftRight = (left: bigint, right: bigint): bigint =>
+    hashN(2, [left, right])
+
+/**
+ * Compute a random BigInt within SNARK_FIELD_SIZE
+ * @return A random BigInt salt.
+ */
+const genRandomNumber = (): bigint => {
```
vimwitch, line 90: Really don't like this rename. Is this implementation any different than `genRandomSalt`?

packages/crypto/src/poseidon.ts
```diff
@@ -0,0 +1,87 @@
+import assert from 'assert'
```
vimwitch, line 1: Where did this file come from/what are we using it for?

packages/crypto/test/IncrementalMerkleTree.test.ts
```diff
@@ -0,0 +1,38 @@
+import { expect } from 'chai'
+import { genRandomNumber, IncrementalMerkleTree } from '../src'
+
+describe('IncrementalMerkleTree', function () {
+    let tree: IncrementalMerkleTree
+    const depth = 4
+
+    it('constructor', async () => {
```
vimwitch, line 8: We use the `zk-kit` implementation for this, we shouldn't test it ourselves. Instead our higher level tests should fail if the tree implementation is broken.

vimwitch (changes requested):
From what I can tell this PR mostly changes some functions.

crypto - `genRandomSalt()` -> `genRandomNum()`
crypto - `IncrementalMerkleTree` constructor arguments
crypto - `SparseMerkleTree.getRootHash()` -> `SparseMerkleTree.root`
smt - `getMerkleProof` -> `createProof`

types - `BigInt` -> `bigint`

I think these changes are more about preference than need. Changing function names and types break any code that depends on this. We should avoid making these types of changes until we have to, and then we should plan and implement it in one go in all of our packages.

The question I'm asking myself here is "are these changes worth the cost?"

vimwitch:
I was also thinking a few days ago that we should almost get rid of the crypto package. We're mostly re-exporting other implementations (which is a good thing). I think it's okay to curate things a bit so people don't have to pull `zk-kit` and `maci-crypto` and other stuff, but we should avoid monkey patching the exports.

packages/crypto/src/IncrementalMerkleTree.ts
vivianjeng: yeah crypto doesn't have to maintain this. Just to make downsteam easier to use

packages/crypto/src/SparseMerkleTree.ts
vivianjeng: Their implementation is really different. I discussed with them but haven't found a solution to use the package. So I just kept the old one.

yeah we can remove `keyv`

packages/crypto/src/crypto.ts
vivianjeng: Because it will download all dependencies we don't need it while installation. 😅
I want to decrease the dependencies and we can only maintain what we need

packages/crypto/src/crypto.ts
vivianjeng: actually no. Yeah we can use the old name

packages/crypto/src/poseidon.ts
vivianjeng: it is from `circomlibjs@0.0.8` 
also don't want to depend on too many packages we don't need 😅

vivianjeng:
New changes has been made:
1. revert `genRandomNumber` to `genRandomSalt`
2. remove `db` as input in `SparseMerkleTree`. but it will be more easily to run out of memory.
    then `SparseMerkleTree` related functions don't need to be `async`

And about `maci-crypto` and `circomlibjs` dependencies, I think we don't need to depend on them to install what we don't need. How do you think? (I am also okay to move the packages back😂)
(I think maybe we (and PSE team) can create a more dependencies-free package and we can import it like `zk-kit/poseidon` without installing other irrelevant packages.)

vimwitch:
Dependency management with nodejs is a bit of a nightmare. We need to look at some actual numbers when we make these decisions. On the `master` branch a clean install with no yarn lock results in `1005M` of `node_modules`. On this branch a clean install makes a `1.1G` `node_modules`. I'm not sure how but the package changes increased the size.

We really want to focus on reducing 2 things, production dependency size (when running servers that depend on Unirep) and the bundle size (in browser). On servers the dependencies can be a bit bigger because the much larger item with be the database file. To help here we should make sure that our `dependencies` and `devDependencies` are accurate. For browsers we want to make sure that we don't bundle multiple versions of packages, and that packages we use are tree-shakeable. For `maci-crypto` they export each variable separately, so anything we don't import will be removed from the bundle automatically. The same is true for `circomlibjs`, the output bundle size should be the same if we copy the code like in this pr or if we import normally. So I think it's fine to revert and just import.

Looking at the dependencies closer, even though we removed `keyv` from our packages we still transitively depend on `web3` which depends on `keyv` D: Still good though because it's not being bundled now. On server the best we can do is make sure that `yarn --production` works (for example this will exclude `keyv` because it's `@ethereum-waffle`->`ganache-core`->`web3`->`keyv`)

vivianjeng:
> I'm not sure how but the package changes increased the size.

Sorry I also some git commit tools (but I removed now). It can be in another PR.
It is 891M now

> On servers the dependencies can be a bit bigger because the much larger item with be the database file.

Yeah maybe I worried too much.

I reverted some changes, so the PR now is 
1. Set `defaultLeaf=0` and `arity=2` in `IncrementalMerkleTree`
2. Use `poseidon` from `circomlibjs`
3. Remove `keyv` from `SparseMerkleTree` and its methods are non-async functions
4. Update packages and dependencies
5. Use the same method names in `IncrementalMerkleTree` and `SparseMerkleTree`
(tree.root, tree.createProof(), tree.verifyProof())
6. Add crypto CI
7. Update README.md

packages/crypto/src/identity.ts
```diff
@@ -1,13 +1,7 @@
-import { BigNumber } from '@ethersproject/bignumber'
-import { randomBytes } from '@ethersproject/random'
-import _sha256 from 'crypto-js/sha256'
-import { poseidon } from 'circomlibjs'
+import { createHash } from 'crypto'
+const { poseidon } = require('circomlibjs')
```
vimwitch, line 2: `require` -> `import`

vimwitch (approved):
LGTM, just need to switch `require` to `import` in a couple places. Let's make sure to squash the commits during merge

Issue #24, opened 2022-05-12
Synchronizer refactor


vimwitch:
I think this will happen in a few stages.

1. Refactor backend to use sqlite wrapped in anondb
2. Pull the sync logic from the backend into the unirep package
  2a. Add event emitter logic for any specific logic for the backend
3. Write a `localStorage` connector for anondb
4. Refactor the frontend to use the unirep synchronizer implementation
  4a. Use `localStorage` by default, and `indexeddb` if available

PR #49, opened 2022-06-12
Synchronizer

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #24

## What is the new behavior?

Circuit config can be imported directly by `@unirep/circuits`. Importing using `@unirep/circuits/config` caused path problems when linked locally.

A base synchronizer class is included in `@unirep/core`

- [ ] Yes
- [x] No

## Other information

It may be easier to review each commit separately, 340ba87 touches a lot of files.

I think we can build tests and docs for this in a different PR. This code is transitively tested by the backend and the docs will probably change as we change the code.


vimwitch:
I've been thinking more about the key management problem and I think we should provide an interface that applications implement. This interface should provide functions for generating proofs and verifying proofs. That way the application can implement whatever logic they need around proving key management.

I did a lightweight version of this ([verification only](https://github.com/Unirep/Unirep/blob/synchronizer/packages/core/src/Synchronizer.ts#L21)) in the synchronizer. [It accepts an object with a `verifyProof` function](https://github.com/Unirep/Unirep/blob/synchronizer/packages/core/src/Synchronizer.ts#L114) that takes the name of the proof, public signals, and proof data. This way we can provide a simple example implementation to help people get started. [This is the backend implementation](https://github.com/Unirep/Unirep-Social-backend/blob/synchronizer-migration/src/daemons/Prover.ts).

vivianjeng:
> That way the application can implement whatever logic they need around proving key management.

Yeah I think so too. Either users can use on-chain verifiers or off-chain snarkjs or other tools.

I think can the Synchronizer be extended to generate `UnirepState` and `UserState`, right?
so we don't need `genUserState`, `genUnirepState`
It will be better if Synchronizer can be extended by these two

I would prefer tests including in this PR as well so that we can test if the Synchronizer works well and it can replace `genUserState`, `genUnirepState`.

vimwitch:
> I think can the Synchronizer be extended to generate UnirepState and UserState, right?

I agree sorta, `UserState` tracks specific data for a certain identity. The Synchronizer shouldn't be tied to an identity. For example, an attester like the backend doesn't need a specific identity.

I think instead the synchronizer can replace the `UnirepState`. The problem is that right now `UnirepState` and `UserState`are really tightly coupled. `UserState` actually gives data to `UnirepState`, which can't happen with the synchronizer; it has to only get data from the blockchain.

The frontend sorta does this now by hooking synchronizer events and then manually updating `UserState`, but ideally it would be the opposite, the `UserState` listens for events emitted by the synchronizer.

image

^ this is how i see data flowing in a refactor.

I think this should definitely happen in a separate pr because it might be backward incompatible + it's just a lot to change.

vivianjeng:
I mean it can replace `genUserState` in `core/src/utils.ts`
https://github.com/Unirep/Unirep/blob/4f4e0d4c1e5bcc301afd10767bd4819668c63256/packages/core/src/utils.ts#L741
Synchronizer can verify events and another object (e.g. `UserSynchronizer`) extends Synchronizer to update the `UserState` object. `UserSynchronizer` takes id as input and update its `userState` if events match
like you said
> The frontend sorta does this now by hooking synchronizer events and then manually updating UserState

---

> The problem is that right now UnirepState and UserStateare really tightly coupled. UserState actually gives data to UnirepState, which can't happen with the synchronizer

I think if `UnirepState` is updated by Synchronizer, `UserState` doesn't need to update `UnirepState`
And if user can get current `UnirepState` from Synchronizer (e.g. current epoch, epoch tree, GST...)
Or `UserState` has to maintain the state itself like current `UnirepState` object (if we remove `UnirepState`)

Now I think I more prefer
~`genUnirepState`~ `Synchronizer` --update--> `UnirepState`
~`genUserState`~ `UserSynchronizer` (extends `Synchronizer`) --updates--> `UserState` (extends `UnirepState`)

`UnirepState` can provide `genGST`, `genEpochTree`, `getAttestations` methods and can be used by users/`UserState`
`UserState` can provide `genCircuitInputs` methods

---

Or maybe just a simple test scripts to test the `Synchronizer`?
e.g. `start()` and try a signup event and check if db updates?
( I think signup event is now less likely to be changed)

vimwitch:
> Now I think I more prefer
genUnirepState Synchronizer --update--> UnirepState
genUserState UserSynchronizer (extends Synchronizer) --updates--> UserState (extends UnirepState)

I think we don't need a `UserSynchronizer` here, the `UserState` can listen to the synchronizer events (instead of a subclass manually calling functions).

> UnirepState can provide genGST, genEpochTree, getAttestations methods and can be used by users/UserState
UserState can provide genCircuitInputs methods

I think we can get rid of the `UnirepState` and just use the synchronizer. It has most of the same variables. The synchronizer can do `genGST`, `genEpochTree`, `getAttestations`

PR #65, opened 2022-07-01
Circleci

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): ci

Github Actions

Issue Number: N/A

## What is the new behavior?

Circle CI

- [ ] Yes
- [x] No


vimwitch:
The ci is failing because it's running all the test cases. In the current `develop` branch the `@unirep/core` test command uses the `&` operator to chain multiple tests. This causes the tests to be executed in the background. Their exit codes are not considered when determining success/failure. [This question](https://stackoverflow.com/questions/9258387/bash-ampersand-operator) covers a similar case.

The relevant command is [here](https://github.com/Unirep/Unirep/blob/develop/packages/core/package.json#L11). Only the results of the synchronizer tests are considered in this command.

This is fixed in #58

.circleci/config.yml
```diff
@@ -0,0 +1,125 @@
+version: 2.1
+workflows:
```
vuvoth, line 2: What is node version we are using? Can we specify node version?

.circleci/config.yml
```diff
@@ -0,0 +1,125 @@
...
+    lint_and_test:
+        jobs:
```
vuvoth, line 4: Can we add branch filter? I think we only need run ci for develop and master.

vuvoth (commented):
Nice PR! I left a few comments.

.circleci/config.yml
vimwitch: By default it's v16 but I switched it to v14. We can specify build matrices to test many versions and change the version more easily if we use docker, but I'm using a machine executor because it's faster with the zk stuff.

.circleci/config.yml
vimwitch: We can add branch filters but I think it's reasonable to test all pushes. That's what we do now, I think we should at least test pull request branches. With circleci we get 10x more free execution time on machines that are 2x as big as what github gives us, might as well use it 😀

vimwitch:
Actually there's something odd happening here. The core tests pass on the develop branch on my local machine, not sure why the ci is broken.

vimwitch:
Not sure why but on the develop branch individual test files fail but the test command doesn't. Removed the parallelism and just used `yarn core test`.

packages/core/package.json
```diff
@@ -8,7 +8,8 @@
         "pre": "node scripts/prepare",
         "build": "tsc",
         "test-cli": "./scripts/testCLI.sh",
-        "test": "yarn synchronizerTests && yarn UnirepStateTests & yarn UserStateTests & yarn integrationTest",
+        "test": "yarn synchronizerTests && yarn UnirepStateTests && yarn UserStateTests && yarn integrationTest",
+        "test-ci": "hardhat test --parallel",
```
vivianjeng, line 12: do we need `test-ci` here?
I think we will change `test` in https://github.com/Unirep/Unirep/pull/58 so maybe we can remove `test-ci`

vivianjeng (approved):
LGTM

PR #69, opened 2022-07-08
Move gen user state to test utils

vimwitch:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

Put `genUserState` in test utils to avoid importing `SQLiteConnector` in downstream dependencies.
Export an ABI json for unirep contracts.
Don't export `defaultProver` from `@unirep/circuits` root.

- [x] Yes
- [ ] No

## Other information

Misc fixes for using unirep packages in frontend and backend.


packages/core/cli/utils.ts
```diff
@@ -71,11 +77,60 @@ const contractExists = async (
     return code.length > 2
 }
 
+/**
+ * Retrieves and parses on-chain Unirep contract data to create an off-chain
+ * representation as a UnirepState object.
+ * @param provider An Ethereum provider
+ * @param address The address of the Unirep contract
+ */
+const genUnirepState = async (
```
vuvoth, line 86: I see we have another genUnirepState in core/test/utils.ts. What is different between them?

packages/core/cli/utils.ts
```diff
@@ -71,11 +77,60 @@ const contractExists = async (
...
+    return synchronizer
+}
+
+/**
+ * This function works mostly the same as genUnirepState,
+ * except that it also updates the user's state during events processing.
+ * @param provider An Ethereum provider
+ * @param address The address of the Unirep contract
+ * @param userIdentity The semaphore identity of the user
+ * @param _userState The stored user state that the function start with
```
vuvoth, line 106: Please update documents comments!

packages/core/test/utils.ts
```diff
@@ -512,6 +513,53 @@ const compareAttestations = (
...
+    return synchronizer
+}
+
+/**
+ * This function works mostly the same as genUnirepState,
+ * except that it also updates the user's state during events processing.
+ * @param provider An Ethereum provider
+ * @param address The address of the Unirep contract
+ * @param userIdentity The semaphore identity of the user
+ * @param _userState The stored user state that the function start with
```
vuvoth, line 542: Please update documents comments.

vuvoth (commented):
I left few comments. Please check it. Thanks <3

packages/core/cli/utils.ts
vimwitch: They're exact copies, I just don't like importing code from the test files in the production code. It's possible to import from the CLI but since it's only in 2 places I figure it's fine. Another options is putting it in a shared utils file in src that isn't exported from index.

We can't export it in the default one because the browser can't use/import `SQLiteConnector`.

vimwitch: So I tried exporting the function and requiring arguments for db and prover, but it seems like you may as well write the whole function at that point. The problem is, we don't want to use `defaultProver` or `SQLiteConnector` so that in a browser, for example, they don't  have to import and then stub `path`, `sqlite`, etc.

packages/contracts/scripts/abi.js
```diff
@@ -0,0 +1,11 @@
+const fs = require('fs')
+const path = require('path')
+
+const ABI = require('../build/artifacts/contracts/Unirep.sol/Unirep.json')
+ABI.bytecode = ''
+ABI.deployedBytecode = ''
```
vivianjeng, line 6: I think we only need `ABI.abi`?

packages/contracts/abi/Unirep.json
```diff
@@ -0,0 +1 @@
...ernalType":"uint256[]","name":"publicSignals","type":"uint256[]"},{"internalType":"uint256[8]","name":"proof","type":"uint256[8]"}],"name":"verifyReputation","outputs":[{"internalType":"bool","name":"","type":"bool"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"uint256[]","name":"publicSignals","type":"uint256[]"},{"internalType":"uint256[8]","name":"proof","type":"uint256[8]"}],"name":"verifyStartTransitionProof","outputs":[{"internalType":"bool","name":"","type":"bool"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"uint256[]","name":"publicSignals","type":"uint256[]"},{"internalType":"uint256[8]","name":"proof","type":"uint256[8]"}],"name":"verifyUserSignUp","outputs":[{"internalType":"bool","name":"","type":"bool"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"uint256[]","name":"publicSignals","type":"uint256[]"},{"internalType":"uint256[8]","name":"proof","type":"uint256[8]"}],"name":"verifyUserStateTransition","outputs":[{"internalType":"bool","name":"","type":"bool"}],"stateMutability":"view","type":"function"}],"bytecode":"","deployedBytecode":"","linkReferences":{},"deployedLinkReferences":{}}
```
vivianjeng, line 1: `abi/`  folder should be included in  `contracts/tsconfig.json`?

vivianjeng (approved):
LGTM
Just few minor comments

packages/contracts/abi/Unirep.json
vimwitch: I don't know if the compiler will pick it up. We don't need to compile it anyway, it's just json.

vuvoth (approved):
LGTM

Issue #66, opened 2022-07-05
Deploy with existing Verifier addresses


vivianjeng:
**Is your feature request related to a problem? Please describe.**
https://github.com/Unirep/Unirep/pull/58#discussion_r913072139
Deploy verifiers every time when deploy a new Unirep.
We can reuse the deployed verifiers if the verifiers are the same

**Describe the solution you'd like**
Deploy contracts with verifier addresses as inputs
If it is `undefined`, then deploy verifiers

**Describe alternatives you've considered**
https://github.com/Unirep/Unirep/pull/58#discussion_r913278761
it would be nice to write a deployer that can retrieve a verifier contract by config. E.g. hash the gst depth, ust depth, epoch tree deptch, etc, then key that to an address of a verifier. Ideally we'd do this with create2 but maybe it's not possible.

ChialiTsai:
Should this one be closed? @vivianjeng

vivianjeng:
not yet 😢

vimwitch:
Closed by #372

PR #73, opened 2022-07-12
Deploy unirep contract with verifiers' addresses as input

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: https://github.com/Unirep/Unirep/issues/66

## What is the new behavior?

- example of new behaviors
  - `npx hardhat deploy:Unirep`
  - `npx hardhat deploy:Verifier --circuit verifyEpochKey`
- test with `run(deploy:Unirep, {attestingFee})` in test scripts

- [ ] Yes
- [x] No


packages/contracts/src/deploy.ts
```diff
@@ -1,132 +0,0 @@
-import { ethers } from 'ethers'
```
vimwitch, line 1: `unirep-social` and the backend rely on this script. Can we do this refactor in a way that isn't breaking? Or at least leave this file in place until pr's are through in the other repos?

packages/contracts/tasks/deploy-unirep.ts
```diff
@@ -0,0 +1,209 @@
+import { Contract } from 'ethers'
+import {
```
vimwitch, line 2: Are we able to use this script outside of a hardhat environment?

vimwitch (changes requested):
Who is this deploy script for? We want other people deploying Unirep to be able to re-use verifiers when possible. I see this working as a `create2` deployment so that the verifier address can be deterministically calculated at deploy time. Then other people that are deploying can check the network for the verifier it needs and use that if it exists.

I don't think this PR does this, right? It allows us to pass a verifier address but you have to know the verifier address (and can't calculate it). Re-using verifiers for tests seems like a bad idea. There's no cost savings but it makes things more complicated.

packages/contracts/src/deploy.ts
vivianjeng: Ideally I think backend should not reply on the script (because the verifiers are related to circuits)
It is also a part of making packages independent of circuits
so I think maybe we can upload circuits/verifiers to a server then the unirep-social can use the desired circuits/verifiers
unirep-social should deploy the verifiers first then deploy `Unirep.sol` in `@unirep/contracts`

but yeah we can still keep this `deploy.ts` until unirep-social is completed.

packages/contracts/tasks/deploy-unirep.ts
vivianjeng: I think that if users download the repo and deploy contract in the repo they have to use the hardhat environment 
and outside the repo they can write their deploy script, set their environment to deploy.
(and they can get `Unirep.sol` from `@unirep/contracts`, get verifiers/circuits from server or build by themselves)

but not 100% sure about this.
does a `deploy.ts` script in `@unirep/contracts` helpful?

Issue #57, opened 2022-06-15
Reputation creation


vimwitch:
Right now we give people 30 rep every epoch. Instead we could use a system where users get a 30 rep subsidy each epoch. The first 30 rep spent is covered by the system. After that it's taken out of the users rep balance. If the user doesn't spend or receive any rep in an epoch their balance is unchanged.

- rep is only created by interactions in the system
- people can't hoard rep by getting airdrops and never spending
- if people create posts/comments just for exchanging rep others can vote and take rep away (vote negatively on the posts/comment for free)

This would also make the case of missing an epoch more clean. Right now if a user doesn't do a UST for an epoch they miss 30 rep for each epoch they skip.

vimwitch:
If a user has negative reputation they will receive an airdrop of 30 rep. e.g. if I'm at -100 rep I will be bumped to -70 rep.

Unclear what will happen if a user is at -15 rep. Maybe they forfeit the remaining 15 rep?

vimwitch:
The first 30 reputation that you spend each epoch is free. So if I have 0 reputation I can still spend reputation without becoming negative.

If I have 5 reputation I can spend the first 30 reputation for free in an epoch, then after that I will be spending my 5 reputation. Or I can wait until next epoch and keep my 5 reputation and get 30 more free rep.

vimwitch:
One way we could implement this is by storing a 2 dimensional mapping of subsidies for users like this:

```
// a mapping of epoch number to aggregate epoch key to subsidy spent
mapping (uint => mapping (uint => uint)) subsidySpent;

// so access like this subsidySpent[epoch][aggregateEpochKey]
```

We want to allow any of the users epoch keys to spend from the subsidy. To do this we can define an aggregate epoch key that is a hashchain of the user epoch keys for a single epoch. e.g. H(H(H(epk0), epk1), epk2). We can then define a zk circuit to prove that an epk belongs to an aggregate epk which accepting the following inputs:
- identity (private)
- epk
- epoch

and outputting:
- aggregate epoch key

--

When creating a post using the subsidy we require the user to submit two proofs:
- a proof spending 0 reputation but proving that their reputation is >= 0
- a proof calculating the aggregate epk from the epk in the spend proof ^

Once we have these two proofs we can check that `subsidySpent[epoch][aggregateEpochKey] + 5 < 30` and then perform the operation.
- If we're creating a post we simple emit a `PostCreated` event
[...]

PR #82, opened 2022-07-19
Prove negative rep

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

#57 

## What is the new behavior?

Adds a proof to show max reputation (e.g. i have less than or equal to X rep)

- [ ] Yes
- [x] No

### DIFF
--- a/packages/circuits/circuits/proveNegativeReputation.circom
+++ b/packages/circuits/circuits/proveNegativeReputation.circom
@@ -1,0 +1,103 @@
+/*
+    Prove:
+        1. if user has a leaf in an existed global state tree
+        2. user state tree has the claimed reputation (pos_rep, neg_rep, graffiti)
+        3. output an epoch key
+*/
+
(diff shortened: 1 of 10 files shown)


vivianjeng:
I think it should be in unirep-social repo
and let unirep-social deploy the verifier

vivianjeng:
I think `@unirep/circuits` only host main circuits that are used in unirep
and actually I want like building circuits with `Circuit` enum iteration
but if there are so many circuits in `Circuit` enum we will build a lot of circuits we don't need in unirep

I think the circuits like negative rep are not that universal as `circomlib` circuits
I prefer to keep the circuits in unirep social first 
and it will be more easily to change the circuit in unirep social (and don't go through `@unirep/circuits` package)

vivianjeng:
I think circuits can still export `*.vkey.json`
and we can upload the parameters on https://www.trusted-setup-pse.org/
so that it can be not only used in verifiers (?!

PR #83, opened 2022-07-19
refactor(circuits): remove unnecessary inputs

vimwitch:
## Pull request checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`yarn build`) was run locally and any changes were pushed
- [x] Lint (`yarn lint --check`) has passed locally and any fixes were made for failures

## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Prove reputation circuit accepts reputation nullifiers as input.

## What is the new behavior?

Prove reputation circuit doesn't accept reputation nullifiers as input.
Remove inputs that can be calculated.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #84, opened 2022-07-20
feat: attest to gst root

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## What is the new behavior?

Allow an attester to submit an attestation with only a gst root. The synchronizer processes the attestation if the GST root exists in the current epoch. This is designed to let attesters use custom zk proofs without Unirep needing to know about them.

- [ ] Yes
- [x] No


packages/contracts/test/Attesting.test.ts
```diff
@@ -395,4 +395,100 @@ describe('Attesting', () => {
             await hardhatEthers.provider.getBalance(unirepContract.address)
         ).to.equal(0)
     })
+
+    it('gst attestation from unregistered attester should fail', async () => {
+        let nonAttester = accounts[5]
```
vuvoth, line 400: I think we should use `const` keyword more than `let`.

packages/core/test/synchronizer/synchronizer.test.ts
```diff
@@ -136,6 +136,97 @@ describe('Synchronizer process events', function () {
...
+        const attestationEvent = new Promise((rs, rj) =>
+            synchronizer.once(GSTAttestationSubmitted, (event) => rs(event))
+        )
+        await synchronizer.unirepContract
+            .connect(accounts[1])
+            .submitGSTAttestation(attestation, epochKey, 0, {
+                value: attestingFee,
+            })
+            .then((t) => t.wait())
+        await attestationEvent
```
vuvoth, line 217: How is `attestationEvent` work?

vuvoth (approved):
LGTM

packages/core/test/synchronizer/synchronizer.test.ts
vimwitch: Here we wait for the synchronizer to process the attestation event. The `attestationEvent` variable is a promise that resolves when the sychronizer processes the event. We just want to make sure the synchronizer picks up the event.

packages/contracts/test/Attesting.test.ts
vimwitch: Definitely, not sure how all these slipped through

vimwitch:
Tests are failing because of #86. A temporary fix is in #83

packages/contracts/contracts/Unirep.sol
```diff
@@ -239,6 +239,38 @@ contract Unirep is IUnirep, zkSNARKHelper, VerifySignature {
         );
     }
 
+    function submitGSTAttestation(
```
vivianjeng, line 242: How about we also overload `submitAttestation`
They are too similar in `_submitAttestation`

or we let it be an optional field in `submitAttestation`
if `gstRoot=0` then check `fromProofIndex` and `toProofIndex`, else the synchronizer checks `gstRoot`

the AttestationSubmitted event can be
```solidity
event AttestationSubmitted(
    uint256 indexed epoch,
    uint256 indexed epochKey,
    address indexed attester,
    AttestationEvent attestationEvent,
    Attestation attestation,
    uint256 toProofIndex,  // optional
    uint256 fromProofIndex, // optional
    uint256 gstRoot // optional
);
```

Or I would prefer we even remove the `fromProofIndex` and `toProofIndex` and `gstRoot` to make the unirep the most lightweight and attesters should be responsible for all reputation they give

Moreover, the hashchain can happen on chain. It is what the previous version unirep did.
(but this is open to discussion)

vivianjeng (changes requested):
I prefer not to create another event and another `submitAttestation` function to do this.
or at least we use a `_checkAttestation` function to do the checks
```solidity
        verifyAttesterSignUp(msg.sender);
        verifyAttesterIndex(msg.sender, attestation.attesterId);
        verifyAttesterFee();

        if (epochKey > maxEpochKey) revert InvalidEpochKey();

        collectedAttestingFee = collectedAttestingFee.add(msg.value);

        // Validate attestation data
        if (!isSNARKField(attestation.posRep))
            revert InvalidSNARKField(AttestationFieldError.POS_REP);

        if (!isSNARKField(attestation.negRep))
            revert InvalidSNARKField(AttestationFieldError.NEG_REP);

        if (!isSNARKField(attestation.graffiti))
            revert InvalidSNARKField(AttestationFieldError.GRAFFITI);
```

packages/contracts/contracts/Unirep.sol
vimwitch: > if gstRoot=0 then check fromProofIndex and toProofIndex, else the synchronizer checks gstRoot

I think we can use a modifier to do the common checks for attestations. I really don't like taking multiple variables and doing lots of if logic to do certain things. Functions should do 1 thing.

> the AttestationSubmitted event can be

I don't think combining the events is a good thing. It makes the attestation processing function very complex off chain. It also requires each event to have fields that are unused.

Also bear in mind `AttestationEvent attestationEvent` is removed in #87. We don't need it because the proof events can identify the type of proof.

> Or I would prefer we even remove the fromProofIndex and toProofIndex and gstRoot to make the unirep the most lightweight and attesters should be responsible for all reputation they give

I don't understand how this would be possible. Would attesters emit proof indexes and gst roots?

vivianjeng:
I think the attesters can verify the GST root by theirselves instead of sending to Unirep.
it is easily to be forged and attester can easily copy a GST root from others
So I think we can just use `submitAttestation(epochKey, attestation)`
and let the attesters verify epk proofs, rep proofs,....and other things

vimwitch:
So an attester wants to give a user 10 reputation if the gst root `X` is valid. An attester can't check this on chain right now. The best they can do is emit a custom event and then conditionally process the attestation. But this breaks other synchronizers and the gst/epoch tree will mismatch between attesters.

Once we have the gst is on chain we can use a raw `submitAttestation` because an attester can verify the gst on chain, but we can't do that right now. We should merge this now because there are refactors to the attestation functions that cleans things up. We can replace `submitGSTAttestation` with `submitAttestation` in the future.

vivianjeng:
yeah so now the synchronizer can process all the attestations as valid attestations
the attester should take responsibility for the attestation he gives 
I think 
1. if the attester is honest
he will verify the gst root / verify proofs before submit attestations
then if we submit gst root to unirep or not will not affect the result
(because the attester wants the valid result and he checks first)
2. if the attester is dishonest
he doesn't verify gst root/ verify proofs before submit attestations
then he can also choose the others' valid gst root and send valid attestations
so it won't affect the result too (the attestation is submitted without verification)

so I think we can use a `submitAttestation(epochKey, attestation)`
and don't need to verify both proofs, gst roots (if we need to customize attester circuits)

vivianjeng:
> 1. if the attester is honest
>    he will verify the gst root / verify proofs before submit attestations

I mean the attester can verify through the off-chain synchronizer like unirep social did before
and I think the attester should do this off-chain

vimwitch:
>  he will verify the gst root / verify proofs before submit attestations

This works for contracts that are admin controlled, but if an attester contract can be called by anyone we can't rely on this.

I think we could have both a gst attestation and a simple `submitAttestation`. Once we put everything on chain we should have just a `submitAttestation`. But we should still merge this because there are cleanups to the smart contract code and we haven't put everything on chain yet.

packages/contracts/contracts/Unirep.sol
```diff
@@ -344,24 +392,21 @@ contract Unirep is IUnirep, zkSNARKHelper, VerifySignature {
...
-        if (publicSignals[0] > maxEpochKey) revert InvalidEpochKey();
-
-        // Add to the cumulated attesting fee
-        collectedAttestingFee = collectedAttestingFee.add(msg.value);
 
         // attestation of airdrop
         Attestation memory attestation;
-        attestation.attesterId = attesters[msg.sender];
+        attestation.attesterId = publicSignals[3];
         attestation.posRep = airdropAmount[msg.sender];
```
vivianjeng, line 402: should we verify `attesters[msg.sender] == publicSignals[3]`?

packages/contracts/contracts/Unirep.sol
```diff
@@ -408,25 +453,19 @@ contract Unirep is IUnirep, zkSNARKHelper, VerifySignature {
...
-        // Add to the cumulated attesting fee
-        collectedAttestingFee = collectedAttestingFee.add(msg.value);
 
         // attestation of spending reputation
         Attestation memory attestation;
-        attestation.attesterId = attesters[msg.sender];
+        attestation.attesterId = publicSignals[maxReputationBudget + 3];
         attestation.negRep = publicSignals[maxReputationBudget + 4];
 
+        assertValidAttestation(msg.sender, attestation, publicSignals[0]);
```
vivianjeng, line 465: also here
`attesters[msg.sender] == maxReputationBudget + 3`

packages/contracts/contracts/Unirep.sol
vimwitch: Yes, but this function is removed in #87

vivianjeng (approved):
I think we can add test scripts for `submitRawAttestation` as well.

packages/contracts/contracts/Unirep.sol
vimwitch: Actually this is verified by [this function](https://github.com/Unirep/Unirep/pull/84/files/bbba5342b873497fb3e090d28bc7b90f857594ee#diff-bd4415896551787d4d50aa8e8bf1cf39f56c05251c9f8fdb9b1746aca509ec02R208).

PR #102, opened 2022-08-10
SMT solidity implementation

### LINKED ISSUE #97: Implement an SMT on chain
zk-kit doesn't include an SMT contract. We may need to write our own implementation.

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Closes #97 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/contracts/contracts/SparseMerkleTree.sol
+++ b/packages/contracts/contracts/SparseMerkleTree.sol
@@ -13,7 +13,7 @@ struct SparseTreeData {
 }
 
 library SparseMerkleTree {
-    uint8 internal constant MAX_DEPTH = 32;
+    uint8 internal constant MAX_DEPTH = 255;
     uint256 internal constant SNARK_SCALAR_FIELD =
         21888242871839275222246405745257275088548364400416034343698204186575808495617;
 
--- a/packages/contracts/contracts/SparseMerkleTree.sol
+++ b/packages/contracts/contracts/SparseMerkleTree.sol
@@ -1,0 +1,80 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {Poseidon2} from './Hash.sol';
+
+struct SparseTreeData {
+    uint256 depth;
+    uint256 root;
+    // depth to zero node
+    mapping(uint256 => uint256) zeroes;
+    // depth to index to leaf
+    mapping(uint256 => mapping(uint256 => uint256)) leaves;
+}
+
+library SparseMerkleTree {
+    uint8 internal constant MAX_DEPTH = 32;
+    uint256 internal constant SNARK_SCALAR_FIELD =
+        21888242871839275222246405745257275088548364400416034343698204186575808495617;
+
+    function init(
+        SparseTreeData storage self,
+        uint256 depth,
+        uint256 _zero
+    ) public {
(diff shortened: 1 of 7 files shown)


packages/contracts/contracts/SparseMerkleTree.sol
```diff
@@ -0,0 +1,80 @@
...
+        require(leaf < SNARK_SCALAR_FIELD);
+        require(index < 2**depth);
+
+        uint256 hash = leaf;
+        uint256 lastLeftElement;
+        uint256 lastRightElement;
+
+        for (uint8 i = 0; i < depth; ) {
+            self.leaves[i][index] = hash;
+            if (index % 2 == 0) {
```
vuvoth, line 58: I think `index & 1` will more efficiency

packages/contracts/contracts/SparseMerkleTree.sol
```diff
@@ -0,0 +1,80 @@
...
+    ) public {
+        uint256 depth = self.depth;
+        require(leaf < SNARK_SCALAR_FIELD);
+        require(index < 2**depth);
+
+        uint256 hash = leaf;
+        uint256 lastLeftElement;
+        uint256 lastRightElement;
+
+        for (uint8 i = 0; i < depth; ) {
```
vuvoth, line 56: why we don't uint256? In my view use u256 will make program more sense.

packages/contracts/test/SMT.ts
```diff
@@ -0,0 +1,97 @@
...
+        poseidon
+    ) as any) {
+        const f = new ethers.ContractFactory(abi, bytecode, accounts[0])
+        const c = await f.deploy()
+        await c.deployed()
+        libraries[`Poseidon${inputCount}`] = c.address
+    }
+    delete libraries['Poseidon3']
+    delete libraries['Poseidon5']
+    const merkleTreeLibFactory = await ethers.getContractFactory(
```
vuvoth, line 19: Could we move deploy SMT lib to beforeEach?

packages/contracts/contracts/SparseMerkleTree.sol
vimwitch: I think the 8 bit integer is cheaper on the stack. The depth can't be more than 32 so it we never need more than 8 bits here.

packages/contracts/test/SMT.ts
vimwitch: Maybe, this is the same problem as before. There's no way to pass variables from `beforeEach` to the test function. I'd rather use this function to deploy than have shared state variables being mutated.

packages/contracts/contracts/SparseMerkleTree.sol
```diff
@@ -0,0 +1,80 @@
...
+    uint256 depth;
+    uint256 root;
+    // depth to zero node
+    mapping(uint256 => uint256) zeroes;
+    // depth to index to leaf
+    mapping(uint256 => mapping(uint256 => uint256)) leaves;
+}
+
+library SparseMerkleTree {
+    uint8 internal constant MAX_DEPTH = 32;
```
vivianjeng, line 16: Can it also be built with a template?
and `64` is better if possible.

Issue #99, opened 2022-08-09
Build GST on chain


ChialiTsai:
When a proof is submitted we should be able to validate the GST root provided. This means updating the GST on every operation, just like the synchronizer.

PR #107, opened 2022-08-12
Onchain epoch tree

### LINKED ISSUE #98: Build epoch tree on chain
Build the epoch tree as attestations are created.

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Closes #98

- [ ] Yes
- [x] No


vimwitch:
@vivianjeng I think this is ready for review. It's a bit blocked by #99, I can't add attestations to the epoch tree without checking the gst. We also need the on chain poseidon functions in this pr for #99.

packages/contracts/contracts/Unirep.sol
```diff
@@ -235,6 +241,38 @@ contract Unirep is IUnirep, zkSNARKHelper, VerifySignature {
             revert InvalidSNARKField(AttestationFieldError.GRAFFITI);
     }
 
+    // increment the hashchain and leave the chain unsealed.
+    // Also store a sealed copy of the hashchain in the epoch tree
+    function storeAttestation(
+        Attestation calldata attestation,
+        uint256 epochKey
+    ) public {
```
vivianjeng, line 249: I think it should be private
After an attestation event comes, it update the attestation hash chains.

packages/contracts/contracts/Unirep.sol
```diff
@@ -235,6 +241,38 @@ contract Unirep is IUnirep, zkSNARKHelper, VerifySignature {
...
+        );
+        // store the latest unsealed hashchain so we can add to it later
+        attestationHashchains[currentEpoch][epochKey] = newHashchain;
+        // then store the sealed hashchain
+        uint256 sealedHashchain = Poseidon2.poseidon([1, newHashchain]);
+        SparseMerkleTree.update(
+            epochTrees[currentEpoch],
+            epochKey,
+            sealedHashchain
+        );
```
vivianjeng, line 273: It should be sealed in the `beginEpochTransition` event.
After the epoch ends, then the `beginEpochTransition` function seals all `hashchains[currentEpoch]` and updates the epoch tree.

vivianjeng (changes requested):

packages/contracts/contracts/Unirep.sol
vimwitch: This is an optimization. In the `beginEpochTransition` function it isn't feasible to seal all known epoch keys - this could be thousands. Instead I seal the hashchain everytime I insert into the epoch tree, but also keep an unsealed version of the hashchain that I can use for future attestations. This way the root of the epoch tree is always accurate (storing unsealed hashchains would yield an inaccurate root i think).

vivianjeng: does it require a circuit change?
In `processAttestations` circuit it will compute the hash chain according to the unsealed hash chain.
I think if it requires a circuit change we can keep it the original version (seal until epoch ends) and then after everything is on-chain then we change the circuit.

vimwitch: I don't think so. We have the unsealed hashchain for each epoch key if we need to verify it. But the final UST circuit outputs a tree root from a sealed hashchain. The synchronizer `genEpochTree` function uses sealed hash chains to build the tree. So when we verify the epoch tree root at the end it should be the same.

vivianjeng: okay I misunderstood the logic
I think it works here.

PR #108, opened 2022-08-12
feat(contracts): getters for config values

### LINKED ISSUE #72: Access config on Unirep contract more easily
Right now we have to do [this](
https://github.com/Unirep/Unirep-Social/blob/8910eb01f20a0eed6978b74744f27eefad79f296/packages/core/contracts/UnirepSocial.sol#L143) to access config variables from outside the Unirep contract.

```
(,,,,uint maxReputationBudget,,,uint attestingFee,,) = unirep.config();
```

This really isn't great. Maybe we could add getters for each individual value?

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Closes #72

- [ ] Yes
- [x] No


packages/contracts/contracts/Unirep.sol
```diff
@@ -748,4 +748,44 @@ contract Unirep is IUnirep, zkSNARKHelper, VerifySignature {
         epochTransitionCompensation[msg.sender] = 0;
         Address.sendValue(payable(msg.sender), amount);
     }
+
+    function globalStateTreeDepth() public view returns (uint8) {
```
vuvoth, line 752: Because we create getters for config values I think we should declare `config` as private data.

vuvoth (approved):

packages/contracts/contracts/Unirep.sol
vimwitch: Ah we still use the `config` variable off chain to load the full config in 1 rpc call. These getters are mostly useful on chain because the config struct is returned as a list instead of a struct in other contracts.

vuvoth: Thanks <3

vivianjeng (approved):
LGTM

PR #126, opened 2022-08-29
Offchain epoch tree

### LINKED ISSUE #123: Build epoch tree offchain
Emit events to build the epoch tree off chain so we can make proofs offchain.

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

- Closes #123 
-
-

- [x] Yes
- [ ] No


packages/contracts/contracts/Unirep.sol
```diff
@@ -60,6 +60,7 @@ contract Unirep is IUnirep, zkSNARKHelper, VerifySignature {
     mapping(uint256 => IncrementalTreeData) public globalStateTree;
     // epoch => root => bool
     mapping(uint256 => mapping(uint256 => bool)) public globalStateTreeRoots;
+    mapping(uint256 => uint256) public epochTreeRoots;
```
vimwitch, line 63: Why do we need this mapping?

vimwitch: We have [this function](https://github.com/Unirep/Unirep/blob/main/packages/contracts/contracts/Unirep.sol#L919) that can be used to load epoch tree roots.

packages/core/src/Synchronizer.ts
```diff
@@ -442,17 +466,46 @@ export class Synchronizer extends EventEmitter {
...
 
+    async epochTreeLeaf(event: ethers.Event, db: TransactionDB) {
+        const epoch = Number(event.topics[1])
+        const leaf = BigInt(event.topics[2]).toString()
+        const index = BigInt(event.topics[3]).toString()
+
+        const exist = await this._db.findOne('EpochTreeLeaf', {
+            where: { epoch, index },
+        })
+        if (exist) {
```
vimwitch, line 488: You can use an upsert operation here:
```
db.upsert('EpochTreeLeaf', {
  where: {
    epoch,
    index,
  },
  update: {
    leaf,
  },
  create: {
    epoch,
    index,
    leaf,
  }
})
```
This will be accelerated by the database when supported.

packages/core/test/synchronizer/epochTree.test.ts
```diff
@@ -0,0 +1,117 @@
...
+            const onchainEpochTree = await unirepContract.epochTrees(epoch)
+            const offchainEpochTree = await synchronizer.genEpochTree(epoch)
+            expect(onchainEpochTree.root.toString()).to.equal(
+                offchainEpochTree.root.toString()
+            )
+        }
+    })
+
+    it('Attest to the same epoch keys should create the same epoch trees onchain and offchain', async () => {
+        const epoch = 1
```
vimwitch, line 77: These tests are comparing sychronizers against themselves. We really need to calculate the expected leaf from an attestation and manually insert it into a tree, then compare that against the synchronizer.

Kinda like in the signup tests where we have an incremental merkle tree and manually update it as we sign users up, then compare the roots against the user state gst root.

vimwitch (changes requested):
Looks good, just needs some small changes

packages/contracts/contracts/Unirep.sol
vivianjeng: I think in the future the `epochTree[epoch]` will be destructed?

packages/core/src/Synchronizer.ts
vivianjeng: It is weird that it will cause error `Error processing event: Error: anondb error: Error: SQLITE_ERROR: near ")": syntax error`

packages/core/src/Synchronizer.ts
```diff
@@ -328,6 +329,23 @@ export class Synchronizer extends EventEmitter {
         return tree
     }
 
+    async genEpochTree(epoch: number): Promise<SparseMerkleTree> {
```
vuvoth, line 332: I think we should use `bigint|number` at here b/c sometime we use get data from smart contract by ethers.js and it return big number I think

packages/contracts/contracts/Unirep.sol
vimwitch: Ah yea, i think it's okay to use the function for now and add this structure once we build that logic

vimwitch, line 63: I still think we don't need this structure. We can just read the values using the existing function.

vivianjeng: [ca8e699](https://github.com/Unirep/Unirep/pull/126/commits/ca8e699b3fdab84da63f4aaf33543b86043b2dfc)

vimwitch (approved):

PR #136, opened 2022-09-16
V2

### LINKED ISSUE #128: Attestation DOS attack
**Describe the attack**
A malicious attester can DOS a user by attesting many times in a single epoch, increasing the proving time of a UST significantly. There is no user input required to signup with an attester, the attester just gives and attestation with `signUp: 1`.

We could partially fix this by requiring that a user submit a zk proof confirming that they want to sign up with a certain attester. This would be verified onchain in the unirep signup function.

### LINKED ISSUE #132: Remove `beginEpochTransition` function
We can remove the `beginEpochTransition` function by implementing a `currentEpoch` function like this.

```ts
// startTime is set as block.timestamp during deployment
function currentEpoch() public view returns (uint) {
  return (block.timestamp - startTime) / epochLength;
}
```

This way the current epoch is automatically calculated as time passes. The only problem is re-initializing the on chain trees for the new epoch. We could make it such that the first attestation in the new epoch handles this logic. In this case the first attestation gas fee would be marginally higher (probably <$0.01).

### LINKED ISSUE #134: Unirep V2
This is a tracking issue proposing a new version of the Unirep protocol.

## Attesters

Each attester is a semaphore group. A user does not sign up globally with Unirep, they sign up with each attester.

Attesters have their own epochs, and each epoch has a `stateTree` and an `epochTree`.

Attesters may set their own `epochLength`.

## State tree

The state tree is an incremental merkle tree with leaves `H(idNullifier, attesterId, epoch, posRep, negRep)`, where `idNullifier` is a user secret. These entries are used to prove the amount of starting rep in the current epoch.

## Reputation

Reputation should consist of only `posRep` and `negRep`. `graffiti` and `signUp` are removed.

The removal of `graffiti` is a result of removing the hashchains from the epoch tree. We can't determine the order in which attestations were received between epoch keys, and thus which `graffiti` is the latest one.

Most uses of `graffiti` can be built outside of the Unirep protocol. e.g. a user claims a username using an epoch key. When the user wants to use the username they prove ownership of the old epoch key.

## Epoch tree

The epoch tree is a sparse merkle tree. Each leaf index is an epoch key and each leaf is the balance received in the current epoch, stored as `H(posRep, negRep)`. An attestation updates the summed value in the leaf.

e.g. if my epoch key is `10294` and I have `posRep: 10` and `negRep: 1` my leaf is `H(10, 1)`. If I receive an attestation with `negRep: 2` my new leaf is `H(10, 3)`.

Only the root of this tree is stored onchain. This tree is replaced every epoch.

## Epoch key

Epoch keys are calculated as `H(idNullifier, attesterId, epoch, nonce) % tree_leaf_count`. Epoch keys are unique between attesters during the same epoch.

e.g. A member of attester A and attester B will have different epoch keys for attester A epoch 1 and attester B epoch 1.

## Attestations

When an attestation is made a ZK proof must be included that outputs the new epoch tree root and leaf. This attestation should sum the previous balance of the leaf and the new attestation.

## User state transition

User state transitions happen per-attester. Because epoch tree leaves are no longer hashchains the logic is much more simple. Each epoch key needs a merkle inclusion proof, then the balance is added to a final sum. The new state tree leaf is calculated using this sum.

The time complexity of a state transition in Unirep scales with the number of attestations. In UnirepV1 it is `O(N)`, in UnirepV1.1 it is **`O(1)`**.

## Spending rep

In UnirepV1 we can only spend reputation using nullifiers. This means if I want to spend 1000 rep I have to generate 1000 nullifiers in ZK and put each onchain.

UnirepV1 also has lag when rep is given. If I receive 10 rep in epoch 1 I cannot prove I have this rep until epoch 2.

UnirepV1.1 solves both of these problems by making the user state transition smaller. In a single ZK proof we can prove the start balance (the state tree leaf) and the balance of each epoch key (which contains the latest attestations).

Let's examine the flow of two spend operations. Assume a user Alice has a start balance of `posRep: 50, negRep: 0` and has not spent anything in this epoch.

- Alice spends 10 rep from epoch key `444`
- The attester takes this proof and creates an attestation of `negRep: 10` to epoch key `444`
- Alice wants to spend another 10 rep
    - Alice proves the start balance from state tree of `posRep: 50, negRep: 0`
    - Alice proves her epoch key balances, all are 0, except for epoch key `444` which has `negRep: 10`
    - Alice sums all values and proves she has `posRep: 50, negRep: 10`

A spend operation internally runs a state transition to calculate the latest balance. The structural change to the epoch tree makes this operation light enough to be possible.

### LINKED ISSUE #142: Fix resource contention for epoch tree root
In #134 we moved the epoch tree offchain by using a zk proof to update a tree root stored on chain. This requires attestations to include a zk proof updating the root. The zk proof must use the current root, and only one proof may reference the current root. This creates resource contention for the current root.

Proposed system: store attestations in an onchain structure then use zk proof(s) to commit the attestations in batches.

We can use a combination of a mapping storing epoch keys keyed to values and a hashchain to process a max of `epoch_key_count`, instead of `attestation_count`. e.g. if 5 epoch keys receive 1000 attestations we only need to process 5 things in zk.

A batch is processed in two steps:
1. A hashchain is made with some number of epoch keys
2. A zk proof is submitted changing the epoch tree root according to a hashchain

This ensures the calldata is constant size (just the hashchain head) and that we only have to process the changed epoch keys, not all attestations.

Below is an psuedocode implementation

```js

// This draft assumes we're in the context of a single attester and epoch
// We need these structures for each epoch and attester

uint epochRoot;

mapping (uint => uint[2]) epochKeyBalances;
uint[] owedEpochKeys;

struct Hashchain {
  uint head;
  uint[] epochKeys;
  bool processed;
}
mapping (uint => Hashchain) hashchains;
uint totalHashchains = 0;
uint processedHashchains = 0;
// user state transition may only happen if (totalHashchains == processedHashchains && owedEpochKeys.length == 0)

event EpochTreeLeaf(
  uint256 indexed epoch,
  uint256 indexed leaf,
  uint256 indexed index
);

function attest(uint epochKey, uint posRep, uint negRep) public {
  if (epochKeyBalance[epochKey][0] == 0 && epochKeyBalance[epochKey][1] == 0)
    owedEpochKeys.push(epochKey);
  epochKeyBalance[epochKey][0] += posRep;
  epochKeyBalance[epochKey][1] += negRep;
}

function buildHashchain(uint _numKeys) public {
  // number of epoch keys to put in the hashchain
  uint numKeys = owedEpochKeys.length < _numKeys ? owedEpochKeys.length : _numKeys;
  uint hashchain = 0;
  uint[] memory epochKeys = new uint[numKeys];
  for (uint8 x = owedEpochKeys.length; x >= 0; --x) {
    uint epochKey = owedEpochKeys[x];
    delete owedEpochKeys[x];
    hashchain = Poseidon4(hashchain, epochKey, epochKeyBalances[epochKey][0], epochKeyBalances[epochKey][1]);
    epochKeys.push(epochKey);
  }
  hashchains[hashchain] = new Hashchain({
    head: hashchain,
    epochKeys: epochKeys
  });
  totalHashchains++;
}

// 3 public signals (old epoch root, hashchain, new epoch root)
function processHashchain(uint calldata publicSignals[], uint calldata proof[]) public {
  Hashchain storage hashchain = hashchains[publicSignals[0]];
  require(hashchain.head != 0);
  require(!hashchain.processed);
  uint oldEpochTreeRoot = publicSignals[1];
  require(epochRoot == oldEpochTreeRoot);
  processedHashchains++;
  hashchain.processed = true;
  for (uint8 x = 0; x < hashchain.epochKeys.length; x++) {
    uint epochKey = hashchain.epochKeys[x];
    emit EpochTreeLeaf(
      currentEpoch,
      Poseidon2(epochKeyBalances[epochKey][0], epochKeyBalances[epochKey][1]),
      epochKey
    );
  }
  epochRoot = publicSignals[2];
}
```

### LINKED ISSUE #143: Migrate circuits to circom 2.0
The circuits use an old version of circom. We should update to the latest.

### LINKED ISSUE #148: Add graffiti to UniRep v1.1
We can add Graffiti values to UniRep v1.1 by storing a timestamp with the graffiti. Then in the circuit we take the graffiti value with the largest timestamp value.

### LINKED ISSUE #151: Epoch key signatures
We need a circuit to prove an epoch key, and accept a 32 byte public signal. This way an epoch key can endorse some arbitrary data.

### LINKED ISSUE #157: Unirep contract proof verification
To verify a proof we need to check the zk proof, and check the roots that were output. In the unirep contract we should have functions that verify the proof _and_ check the root outputs and assert that they are valid.

This difference should be noted in the docs. The verifiers can be used directly, but the caller should be aware that the roots may be invalid or non-existent.

### LINKED ISSUE #167: Autoselect event handler based on function name
In the `Synchronizer` we manually bind events to handler functions [here](https://github.com/Unirep/Unirep/blob/epoch-fix/packages/core/src/Synchronizer.ts#L487). When an attester creates a subclass they have to override this function (and some others) mapping their own contract events to their own functions.

Instead we should accept a list of contract addresses keyed to event names like this:
```js
{
  '0xabababababababababababababababababababab': [
    'SectionSubmitted',
    'SectionVote'
  ]
}
```

This way we can support observing multiple events from multiple contracts.

--

Events should be mapping to functions by converting them to lower camelCase. e.g. `SectionSubmitted` invokes `this.sectionSubmitted`. This function should be invoked with (at least) the raw event object, and the transaction database object.

At initialization time if the expected function doesn't exist on the object a warning should be printed. e.g. `this.sectionSubmitted === undefined`. If an event is received with no corresponding handler an error should be thrown.

At initialization time if there are two colliding event names an error should be thrown.

### LINKED ISSUE #173: Store code coverage reports
We should add a build badge to our readmes that shows the code coverage for our packages. This badge should link to the coverage report so we can easily track coverage changes over time. See the badge on [this](https://github.com/vimwitch/anondb) readme. That repo stores the report using web3.storage and the latency is pretty bad. We could build an alternative solution using cloudflare instead.

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Closes #134 
Closes #132 
Closes #128 
Closes #143 
Closes #142 
Closes #148 
Closes #151
Closes #157
Closes #173
Closes #167 

- [x] Yes
- [ ] No


vimwitch:
Needs more tests in `circuits`/`core` packages

packages/crypto/src/crypto.ts
```diff
@@ -23,3 +23,45 @@ export const hashLeftRight = (input1: any, input2: any) =>
...
+    return hash3([BigInt(attesterId), BigInt(epoch), identityNullifier as any])
+}
+
+export const genStateTreeLeaf = (
+    idNullifier: bigint,
+    attesterId: bigint | string,
+    epoch: bigint | number,
+    posRep: bigint | number,
+    negRep: bigint | number
+): BigInt => {
```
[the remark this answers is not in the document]
vivianjeng: I think we can either merge https://github.com/Unirep/Unirep/pull/140 first
Or directly change the type in the PR since there are many differences from #140

packages/circuits/config/index.ts
```diff
@@ -7,23 +9,13 @@ export const proveReputationCircuitPath =
...
     '../zksnarkBuild/proveUserSignUp_main.circom'
 
-export const startTransitionCircuitPath =
-    '../zksnarkBuild/startTransition_main.circom'
-
-export const processAttestationsCircuitPath =
-    '../zksnarkBuild/processAttestations_main.circom'
-
-export const userStateTransitionCircuitPath =
-    '../zksnarkBuild/userStateTransition_main.circom'
```
vivianjeng, line 17: These can also be removed?

vivianjeng (commented):
The protocol looks good to me 👍🏻
but I think we should also add `Synchronizer` and `UserState` tests to test all functions?
like `Synchronizer` should catch all events and generate correct state
and `UserState` should generate proofs correctly.

packages/crypto/src/crypto.ts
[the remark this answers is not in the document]
vimwitch: Yea I think we can make divergent histories here. We put #140 on top of the v1.0.0 tag, and main is a separate history.

packages/circuits/circuits/epochTransition.circom
```diff
@@ -0,0 +1,122 @@
+include "../../../node_modules/circomlib/circuits/poseidon.circom";
```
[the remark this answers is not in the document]
vimwitch: 0ae24b789c8a26539ee6b2b563d05ad64405d466

packages/circuits/config/index.ts
vimwitch: cfd4ed12797ab5c78bf59d8407a6bd33a8e9264e

packages/crypto/src/crypto.ts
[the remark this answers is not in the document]
vimwitch: f909e5b9611513963dbb65ce298e161e868bf56f

packages/contracts/test/attestation.test.ts
```diff
@@ -0,0 +1,272 @@
...
+            r.proof,
+            defaultProver
+        )
+        const _proof = [...proof]
+        _proof[0] = BigInt(_proof[0].toString()) + BigInt(1)
+        await expect(
+            unirepContract
+                .connect(accounts[1])
+                .submitAttestation(0, publicSignals, _proof)
+        ).to.be.reverted
```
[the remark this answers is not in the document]
vimwitch: Ah yea, there's no revert reason in this case. If the proof values are invalid the verifier contract reverts without a reason.

packages/contracts/test/epochTransition.test.ts
```diff
@@ -0,0 +1,584 @@
...
+            r.proof,
+            defaultProver
+        )
+        const _proof = [...proof]
+        _proof[0] = BigInt(proof[0].toString()) + BigInt(1)
+        await expect(
+            unirepContract
+                .connect(accounts[1])
+                .epochTransition(publicSignals, _proof)
+        ).to.be.reverted
```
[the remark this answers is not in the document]
vimwitch: Same for this and the other.

packages/contracts/test/attestation.test.ts
```diff
@@ -0,0 +1,272 @@
...
+        unirepContract = await deployUnirep(accounts[0])
+    })
+
+    it('should have the correct config value', async () => {
+        const config = await unirepContract.config()
+        expect(NUM_EPOCH_KEY_NONCE_PER_EPOCH).equal(
+            config.numEpochKeyNoncePerEpoch
+        )
+        expect(EPOCH_TREE_DEPTH).equal(config.epochTreeDepth)
+        expect(GLOBAL_STATE_TREE_DEPTH).equal(config.globalStateTreeDepth)
```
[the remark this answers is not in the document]
vimwitch: 45611e98b5859f24a13acc03a757695277d990b4

packages/core/src/UserState.ts
```diff
@@ -360,545 +290,185 @@ export default class UserState extends Synchronizer {
...
-    ): Promise<EpochKeyProof> => {
-        this._checkEpkNonce(epochKeyNonce)
-        const epoch = await this.latestTransitionedEpoch()
-        const leafIndex = await this.latestGSTLeafIndex()
-        const epochKey = genEpochKey(
-            this.id.identityNullifier,
-            epoch,
-            epochKeyNonce,
-            this.settings.epochTreeDepth
+    public genAttestationProof = async (
```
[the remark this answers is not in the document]
vimwitch: Ah it uses `getRepByEpochKey` which is in user state. We could put that in Synchronizer too, but i think this will change a lot when we do #142. Maybe we can do a better solution then.

packages/crypto/src/crypto.ts
```diff
@@ -23,3 +23,45 @@ export const hashLeftRight = (input1: any, input2: any) =>
...
+    const epochKey = hash4([
+        identityNullifier as any,
+        BigInt(attesterId),
+        epoch,
+        BigInt(nonce),
+    ]).valueOf()
+    // Adjust epoch key size according to epoch tree depth
+    const epochKeyModed = epochKey % BigInt(maxEpochKey)
+    return epochKeyModed
+}
```
vivianjeng, line 43: I think this function should be in `@unirep/core` because it is related to the protocol
the `@unirep/crypto` should be more general crypto functions

and it can be

```ts
export const genEpochKey = (
    identityNullifier: bigint,
    attesterId: bigint | string,
    epoch: bigint | number,
    nonce: bigint | number,
    maxEpochKey: bigint | number
): bigint => {
    const epochKey = hash4([
        identityNullifier,
        BigInt(attesterId),
        epoch,
        BigInt(nonce),
    ])
    // Adjust epoch key size according to epoch tree depth
    const epochKeyModed = epochKey % BigInt(maxEpochKey)
    return epochKeyModed
}
```

packages/crypto/src/crypto.ts
```diff
@@ -23,3 +23,45 @@ export const hashLeftRight = (input1: any, input2: any) =>
...
+    negRep: bigint | number
+): bigint => {
+    return hash5([
+        idNullifier,
+        BigInt(attesterId),
+        BigInt(epoch),
+        BigInt(posRep),
+        BigInt(negRep),
+    ])
+}
```
vivianjeng, line 67: these two functions are also more protocol related so that it should be in `@unirep/core`

vivianjeng (changes requested):

packages/crypto/src/crypto.ts
vimwitch: It's nice having these functions in the crypto package because a client can calculate an epoch key or nullifier without needing to import core. We use these functions in the circuits/contracts packages so if we put `genEpochKey` in core we have a circular dependence.

The `@unirep/crypto` package is part of the unirep protocol I think, we're not writing crypto functions (e.g. implementing poseidon), these are just functions that are helpful when working with the protocol.

vivianjeng: `genEpochKey` is only used in test scripts in circuits and contracts
it is not a core function in circuits and contracts so it won't cause circular dependence

but it we put it in different package it misleads developers and `@unirep/crypto` should not handle what we **define** (e.g. epochKey, epochKeyNullifier) in the protocol

vimwitch: I don't see how it misleads developers, the functions are named according to our protocol.

> @unirep/crypto should not handle what we define

I guess I don't understand this, why shouldn't we put our key calculation functions in our crypto package?

> it is not a core function in circuits and contracts so it won't cause circular dependence

ehhh these functions were defined in multiple places in the past. Maybe it's not a breaking circular dependence, but it's certainly a code smell imo

vivianjeng: Here is what we **define** in unirep protocol
https://docs.unirep.io/protocol/glossary

and hash, poseidon, SparseMerkleTree,... are not
and they are helpers of unirep protocol

yeah I agree they are defined in many places is annoying.
Or maybe we can rename it to `@unirep/utils` to handle these like protocol related functions but they are used in many packages

vimwitch: Right I get that right now it's all completely generic structures, I don't see why it's bad to include unirep specific stuff in this package. It's under our namespace `@unirep` and it doesn't affect the other exports.

vivianjeng: but maybe `@unriep/crypto` is confusing to maintain all these functions.
So how do you think `@unirep/utils`?
I think it will affect like unirep social
Since it uses `genEpochKey()` functions
and previously it is in `@unriep/core`
https://github.com/Unirep/Unirep-Social/blob/40ff6ea26f514718409807210bd913655655c755/packages/frontend/src/context/User.ts#L7

vimwitch: > but maybe @unriep/crypto is confusing to maintain all these functions.

I think it's okay as long as it's clear in the documentation.

> I think it will affect like unirep social

Yes there will be breaking changes for downstream applications. We can make it backward compatible by re-exporting it from `@unirep/core`.

Actually looking at [semver](https://semver.org/) maybe we should be calling this v2 because it has other breaking changes.

vimwitch: >  So how do you think @unirep/utils?

I really don't like this. What are utils? Hash functions are utils, making an epoch key is too, maybe a tree structure is a util.

Using `utils` anywhere is a code smell to me

vivianjeng: > Yes there will be breaking changes for downstream applications. We can make it backward compatible by re-exporting it from @unirep/core.

I don't think in the breaking changes we don't really care about backward compatible. We will anyway change unirep social according to `v1.1`

> Actually looking at [semver](https://semver.org/) maybe we should be calling this v2 because it has other breaking changes.

I think it is okay to call it `v2`

> I really don't like this. What are utils? Hash functions are utils, making an epoch key is too, maybe a tree structure is a util.

yes they are
I think it will be better than put it all in `@unirep/crypto` since some of them are not crypto related
or hash function in `@unirep/crypto` and another package `@unirep/utils`

vimwitch: > I think it is okay to call it v2

👍

>  I think it will be better than put it all in @unirep/crypto since some of them are not crypto related

But something like `genEpochKey` _is_ crypto related. It calculates a specific hash using some inputs. It's like calculating a public key in tls or ssh or semaphore. We put it in the crypto package because it's part of the unirep cypto-protocol.

We could technically break these 4 functions into a separate package, but is that worth the complexity? Both in terms of managing packages, and users figuring out what the different packages are?

packages/core/src/UserState.ts
```diff
@@ -360,545 +289,192 @@ export default class UserState extends Synchronizer {
...
-     * In the circuit, it will compute the reputation nullifiers of the given nonce. If the reputation nullifier is used
-     * to spend reputation, it cannot be spent again.
      * @returns The reputation proof of type `ReputationProof`.
      */
     public genProveReputationProof = async (
-        attesterId: BigInt,
         epkNonce: number,
-        minRep?: number,
-        proveGraffiti?: BigInt,
-        graffitiPreImage?: BigInt,
```
vivianjeng, line 824: should prove the graffiti preimage

packages/core/src/UserState.ts
```diff
@@ -360,545 +289,192 @@ export default class UserState extends Synchronizer {
...
-            sign_up: signUp,
-            UST_path_elements: USTPathElements,
-            rep_nullifiers_amount: spendAmount,
-            start_rep_nonce: nonceStarter,
+            graffiti,
+            timestamp,
             min_rep: minRep === undefined ? 0 : minRep,
-            prove_graffiti: proveGraffiti === undefined ? 0 : proveGraffiti,
-            graffiti_pre_image:
-                graffitiPreImage === undefined ? 0 : graffitiPreImage,
```
vivianjeng, line 889: should prove the graffiti preimage

vivianjeng (changes requested):

packages/core/src/UserState.ts
vivianjeng: [27a46fc](https://github.com/Unirep/Unirep/pull/161/commits/27a46fcad3bba01436b2647d28825e0ee1f57143) fixed

packages/crypto/src/crypto.ts
vivianjeng: For me, `genEpochKey` is not a generally known crypto related function (like `SparseMerkleTree` or `hash`). It is like a specific function used in unirep. 

> We could technically break these 4 functions into a separate package, but is that worth the complexity? Both in terms of managing packages, and users figuring out what the different packages are?

Yeah so I suggest them to be in `@unirep/utils` to be easily found them all
and they don't have to be crypto related

vivianjeng: like `ethers` also put hash in `utils`
https://docs.ethers.io/v5/api/utils/hashing/

vimwitch: But notice that ethers doesn't have a `utils` _and_ a `crypto` package. I'd be fine with changing the entire `crypto` package to be `utils`. I just don't want to maintain many packages that have few functions.

vivianjeng: I agree with that the entire `crypto` to be `utils` 👍🏻

vimwitch: ok, documented in #166

packages/core/src/Synchronizer.ts
```diff
@@ -396,211 +543,161 @@ export class Synchronizer extends EventEmitter {
...
-            hash: attestation.hash().toString(),
+            epoch,
+            epochKey,
+            index,
+            attesterId,
+            posRep,
+            negRep,
+            graffiti: decodedData.graffiti.toString(),
+            timestamp: decodedData.timestamp.toString(),
+            hash: hash2([posRep, negRep]).toString(),
```
vivianjeng, line 650: do we need this?
or it should be `hash4([posRep, negRep, graffiti, timestamp])`

vimwitch: Related #200

PR #155, opened 2022-10-23
V2 verifiers

### LINKED ISSUE #151: Epoch key signatures

### LINKED ISSUE #157: Unirep contract proof verification

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Closes #151 
Closes #157 

## What is the new behavior?

- Document verifiers, change argument ordering to be consistent with other functions.
- Adds hash value to epoch key proof.

- [ ] Yes
- [ ] No


packages/docs/docs/contracts-api/unirep-sol.md
```diff
@@ -103,6 +103,52 @@ function attesterEpochRemainingTime(
...
+A contract address for an epoch key proof verifier. See [IVerifier](/docs/contracts-api/iverifier-sol) for more info.
+
+```sol
+IVerifier public immutable epochKeyVerifier;
+```
+
+Example use:
+
+```sol
+bool valid = unirep.epochKeyVerifier.verifyProof(publicSignals, proof);
```
vivianjeng, line 117: I think we should also let users verify proofs through contract
either
1. create a function in the `Unirep.sol` to let user call verifier: https://github.com/Unirep/Unirep/blob/7945d2e7d95981e489af0d231f8734a11db94a50/packages/contracts/contracts/Unirep.sol#L439
2. export `IVerifier` abi so that user can reconstruct `const verifier = new Contract(address, abi)`

vivianjeng (changes requested):

packages/docs/docs/contracts-api/unirep-sol.md
vimwitch: I disagree, the user would be better off calling the verifier manually. If we implement verifier functions in the unirep contract it should verify the proof AND verify the existence of the state root for the attester+epoch.

We don't need to check `isSNARKField` ourselves, the verifier contract checks this already.

vivianjeng: yeah so if they want to call the verifier they should know the verifier abi
they should also check the state root for the attester+epoch by theirselves
it depends on their applications

vivianjeng: the `verifyEpochKeyProof ` is to make sure the user uses the same verification key as `Unirep.sol`
and it performs like `Prover.verifyProof()`

vimwitch: >  they should also check the state root for the attester+epoch by theirselves

I think this is unreasonable. If we provide a function `verifyEpochKey` in the unirep contract, it's extremely misleading if it doesn't verify everything.

> the verifyEpochKeyProof is to make sure the user uses the same verification key as Unirep.sol

If an external contract does `epochKeyVerifier.verifyProof()` they're using the same verification key right?

vivianjeng: > I think this is unreasonable. If we provide a function verifyEpochKeyProof in the unirep contract, it's extremely misleading if it doesn't verify everything.

It provides `Prover.verifyProof()` without user having `snarkjs` and `vkey`
On second thought, I agree that `verifyEpochKeyProof ` can provide state root check 

> If an external contract does epochKeyVerifier.verifyProof() they're using the same verification key right?

yes I think so

packages/contracts/contracts/Unirep.sol
```diff
@@ -388,6 +388,43 @@ contract Unirep is IUnirep, VerifySignature {
...
 
+    function verifyEpochKeyProof(
+        uint256[] memory publicSignals,
+        uint256[8] memory proof
+    ) public returns (bool) {
+        bool valid = epochKeyVerifier.verifyProof(publicSignals, proof);
+        // short circuit if the proof is invalid
+        if (!valid) return false;
+        require(publicSignals[3] < type(uint160).max);
+        require(publicSignals[0] < maxEpochKey);
```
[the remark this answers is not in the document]
vivianjeng: I also think we should use error message to handle the check.
e.g.
```solidity
if (publicSignals[0] >= maxEpochKey) revert InvalidEpochKey();
```

packages/contracts/contracts/Unirep.sol
```diff
@@ -388,6 +388,43 @@ contract Unirep is IUnirep, VerifySignature {
...
+        updateEpochIfNeeded(uint160(publicSignals[3]));
+        AttesterData storage attester = attesters[uint160(publicSignals[3])];
+        // epoch check
+        if (publicSignals[2] > attester.currentEpoch) return false;
+        // state tree root check
+        if (!attester.stateTreeRoots[publicSignals[2]][publicSignals[1]])
+            return false;
+        return true;
+    }
+
```
vivianjeng, line 427: I think `verifyUSTProof`, `verifySignUpProof`, `verifyAggregateEpochKeysProof` can also be functions in `Unirep.sol`
the public signals can also be verified in these functions

packages/docs/docs/contracts-api/unirep-sol.md
```diff
@@ -103,10 +103,36 @@ function attesterEpochRemainingTime(
...
+    uint256[8] memory proof
+) public returns (bool);
+```
+
 ## epochKeyVerifier
 
 A contract address for an epoch key proof verifier. See [IVerifier](/docs/contracts-api/iverifier-sol) for more info.
 
+:::danger
+Using the verifier directly does not validate the output state root, attester id, or epoch. Prefer the [`verifyEpochKeyProof`](#verifyepochkeyproof) function unless you know what you are doing.
```
vivianjeng, line 141: the updated `verifyEpochKeyProof` checks the state root?

vivianjeng (changes requested):

packages/docs/docs/contracts-api/unirep-sol.md
vimwitch: Yes

packages/contracts/contracts/Unirep.sol
vimwitch: Yea, I think we can do this as a part of #157

packages/circuits/test/verifyEpochKey.test.ts
```diff
@@ -61,6 +61,59 @@ describe('Verify Epoch Key circuits', function () {
...
+                negRep,
+                graffiti,
+                timestamp,
+                data,
+            })
+            const { isValid, publicSignals } = await genProofAndVerify(
+                Circuit.verifyEpochKey,
+                circuitInputs
+            )
+            expect(isValid).to.be.true
```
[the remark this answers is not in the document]
vimwitch: Can you explain a bit more? I'm not sure how this test case is different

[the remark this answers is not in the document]
vivianjeng: I mean the change of `publicSignal[4]` (data) 
`prover.verifyProof(_publicSignals, proof)` will throw `false`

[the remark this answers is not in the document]
vimwitch: This seems like we're just testing the proof system? Changing a public signal should always break the proof.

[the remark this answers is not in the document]
vivianjeng: okay what I remembered is that public input should also set constraints
https://github.com/tornadocash/tornado-core/blob/1ef6a263ac6a0e476d063fcb269a9df65a1bd56a/circuits/withdraw.circom#L61
but maybe I was wrong.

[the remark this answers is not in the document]
vimwitch: Ah ok, yes we should test for this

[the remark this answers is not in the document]
vimwitch: [2050b16](https://github.com/Unirep/Unirep/pull/155/commits/2050b16ea0c685e352cf8aef2f7e6b28c6a4d95f)

vimwitch:
@vivianjeng I don't think we should include verify functions for user state transition or aggregate epoch keys. For UST there are lots of internal state things we check, like making sure all epoch keys are processed. If we return a simple boolean for validity it's hard to know why a proof isn't valid.

UST + aggregate epoch key proofs shouldn't be used outside the protocol. Can you think of cases where another contract would want to verify these proofs manually?

vivianjeng:
> UST + aggregate epoch key proofs shouldn't be used outside the protocol. Can you think of cases where another contract would want to verify these proofs manually?

I am thinking about like unirep social will relay transactions for users they might want to verify proof before queueing the transaction. And verifying through smart contract the server doesn't have to know the `vkey` itself and doesn't require `snarkjs` (just `ethers` to call the function). So we can consider what states are the same **before and after** sending a UST proof. e.g. `stateTreeRoot` of an `epoch` should exist before and after sending a UST proof.

but on-chain attester it might not need these functions because `Unirep.sol` will revert if invalid transactions are sent.

Does it make sense?

vimwitch:
Kinda, but an offchain relay could just estimate the `userStateTransition` function to determine if the proof is fully valid, right? If the relay doesn't have the vkey, then they probably aren't going to call a specific function to verify the proof, they're probably just naively broadcasting transactions. e.g. accept a `to` address and `calldata`. In which case they will know if the proof is valid based on whether gas estimation fails.

vivianjeng:
> If the relay doesn't have the vkey, then they probably aren't going to call a specific function to verify the proof

So I think there can be a specific function in contract to verify the proof
Yeah I also agree users can estimate the `userStateTransition` function to see if the proof is correct or not
or we can keep it like the current PR and to see if we need the specific function in the future.

vivianjeng (approved):

PR #165, opened 2022-10-25
fix: epoch calculation

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Calculates the current epoch offchain using the start timestamp/epoch length. Fixes some problems with test eth nodes not reporting correct timestamp values.

Refactors user state proof functions to accept objects with options instead of many arguments.

- [ ] Yes
- [ ] No


packages/core/src/Synchronizer.ts
```diff
@@ -86,16 +89,13 @@ export class Synchronizer extends EventEmitter {
...
         })
-        if (epochs.length > 1) {
-            throw new Error('Multiple unsealed epochs')
-        }
         this.defaultStateTreeLeaf = BigInt(0)
         this.stateTree = new IncrementalMerkleTree(
             this.settings.stateTreeDepth,
             this.defaultStateTreeLeaf
         )
         // if it's a new sync, start with epoch 1
```
vivianjeng, line 97: description can be updated as well

packages/core/src/Synchronizer.ts
```diff
@@ -306,17 +309,17 @@ export class Synchronizer extends EventEmitter {
...
             )
         }
     }
 
-    protected async _checkValidEpoch(epoch: number) {
-        const currentEpoch = await this.loadCurrentEpoch()
-        if (epoch > Number(currentEpoch)) {
+    protected async _checkValidEpoch(epoch: bigint | number) {
+        const currentEpoch = this.calcCurrentEpoch()
+        if (BigInt(epoch) > currentEpoch) {
```
vivianjeng, line 322: There is a
https://github.com/Unirep/Unirep/blob/f15c0cb0290dd1817487d6312b922a0e2019a73a/packages/core/src/Synchronizer.ts#L612
in `attestationEvent(event: ethers.Event, db: TransactionDB)`
I think if `_checkCurrentEpoch()` means to get the epoch on-chain (not the db stored epoch)
then the `await this._checkCurrentEpoch(_epoch)` can be changed to `await this._checkValidEpoch(_epoch)` or we should also keep the `readCurrentEpoch` to read the current db processed epoch.

Or after epoch ends, the attestation event will not be processed correctly (the synchronizer will break)

vivianjeng (changes requested):

packages/core/src/Synchronizer.ts
vimwitch: What do you mean? This error no longer exists, it's expected that the client has multiple unsealed epochs. This happens if there is no activity for multiple epochs. The next action will trigger an epoch transition across multiple epochs.

The error is completely removed in this PR.

packages/core/src/Synchronizer.ts
vimwitch: Right, added this logic back in here: [52cfb42](https://github.com/Unirep/Unirep/pull/165/commits/52cfb42b99f7b16a2cc88c7b89bfcf933b9113ec)

The `_checkCurrentEpoch` function is only used in a single place. I removed it because the name is ambiguous.

packages/core/src/Synchronizer.ts
vivianjeng: I mean this description line
https://github.com/Unirep/Unirep/blob/52cfb42b99f7b16a2cc88c7b89bfcf933b9113ec/packages/core/src/Synchronizer.ts#L97
Since the epoch updates from `0`

vimwitch: ok [72ce291](https://github.com/Unirep/Unirep/pull/165/commits/72ce29162c972b3231367d858c4bad87e45460de)

vivianjeng (approved):

Issue #163, opened 2022-10-24
V1.1 contract test


vivianjeng:
Build `@unirep/contracts` tests to cover all functions in `Unirep.sol`

vivianjeng:
userSignUp
- [ ] success
- [ ] revert AttesterIdNotMatch()
- [ ] revert InvalidProof()
- [ ] revert AttesterNotSignUp()
- [ ] revert UserAlreadySignedUp()
- [ ] revert EpochNotMatch()
- [ ] state tree updates as expected
- [ ] semaphoreGroup updates as expected

attesterSignUp
- [ ] success
- [ ] revert AttesterAlreadySignUp()

attesterSignUpViaRelayer
- [ ] success
- [ ] revert InvalidSignature()
- [ ] revert AttesterAlreadySignUp()

submitAttestation
- [ ] success
- [ ] revert EpochNotMatch()
- [ ] revert InvalidEpochKey()

buildHashchain
- [ ] success
- [ ] revert when no attestations

processHashchain
- [ ] success
- [ ] revert InvalidProof()
- [ ] revert when `attesterId == 0`
- [ ] revert HashchainInvalid()
- [ ] revert InvalidEpochTreeRoot()

userStateTransition
- [ ] success
- [ ] revert InvalidProof()
- [ ] revert AttesterInvalid()
- [ ] revert NullifierAlreadyUsed()
- [ ] revert EpochNotMatch()
- [ ] revert HashchainNotProcessed()
- [ ] revert InvalidEpochTreeRoot()
- [ ] revert InvalidStateTreeRoot()
- [ ] state tree updates as expected

updateEpochIfNeeded
- [ ] success
- [ ] revert AttesterNotSignUp()
- [ ] epoch tree init correctly
- [ ] state tree init correctly

attesterCurrentEpoch
- [ ] success
- [ ] revert AttesterNotSignUp()

attesterEpochRemainingTime
- [ ] success
- [ ] revert AttesterNotSignUp()

attesterHashchainTotalCount
- [ ] success
- [ ] revert AttesterNotSignUp()

attesterHashchainProcessedCount
- [ ] sucess
- [ ] revert AttesterNotSignUp()

attesterHashchain
- [ ]  sucess
- [ ] revert AttesterNotSignUp()

attesterEpochLength
- [ ] success 
- [ ] revert AttesterNotSignUp()

attesterStateTreeRootExists
- [ ] success 
- [ ] revert AttesterNotSignUp()

[...]

PR #187, opened 2022-10-30
V2 contract test

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #163 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


packages/contracts/contracts/Unirep.sol
```diff
@@ -308,7 +309,6 @@ contract Unirep is IUnirep, VerifySignature {
         // Verify the proof
         if (!userStateTransitionVerifier.verifyProof(proof, publicSignals))
             revert InvalidProof();
-        if (publicSignals[5] >= type(uint160).max) revert AttesterInvalid();
```
vivianjeng, line 311: `AttesterInvalid` is removed here because `AttesterNotSignUp` will handle the same situation

packages/contracts/contracts/Unirep.sol
```diff
@@ -392,13 +391,11 @@ contract Unirep is IUnirep, VerifySignature {
     function verifyEpochKeyProof(
         uint256[] memory publicSignals,
         uint256[8] memory proof
-    ) public returns (bool) {
+    ) public view returns (bool) {
```
vivianjeng, line 394: Turn this to a `view` function because I think it might duplicated if user verify proof and submit attestations
And a `view` function might be more useful if the backend or frontend wants to verify proofs and they don't have to send another transactions

vivianjeng: Okay I understand
so maybe in this case it doesn't have to return a `true` or `false`, it can revert the transaction?
I don't know if there is any case that a `false` proof can execute a valid transaction.
and it will be better to be tested 
(since `ethers` might not get the return value of a transaction?https://ethereum.stackexchange.com/questions/88119/i-see-no-way-to-obtain-the-return-value-of-a-non-view-function-ethers-js)

> verify the roots using the offchain data

yeah I agree with this

packages/contracts/test/userStateTransition.test.ts
```diff
@@ -509,14 +553,13 @@ describe('User State Transition', function () {
...
         stateTree.insert(leaf)
-        const snapshot = await ethers.provider.send('evm_snapshot', [])
         const epochKeys = Array(NUM_EPOCH_KEY_NONCE_PER_EPOCH)
             .fill(null)
             .map((_, i) =>
                 genEpochKey(
                     id.identityNullifier,
                     BigInt(attester.address),
-                    fromEpoch, // from epoch
+                    0, // from epoch
```
vimwitch, line 562: there's a `fromEpoch` above on line 553, we can remove that or use it here?

packages/contracts/test/userStateTransition.test.ts
```diff
@@ -572,81 +617,112 @@ describe('User State Transition', function () {
...
+        const snapshot = await ethers.provider.send('evm_snapshot', [])
+        await ethers.provider.send('evm_increaseTime', [EPOCH_LENGTH])
+        await ethers.provider.send('evm_mine', [])
+        const startEpoch = (
+            await unirepContract.attesterCurrentEpoch(attester.address)
+        ).toNumber()
+        for (let epoch = startEpoch; epoch < startEpoch + epochs; epoch++) {
+            _stateTree[epoch] = new IncrementalMerkleTree(STATE_TREE_DEPTH)
+            for (let i = 0; i < users; i++) {
+                console.log(`epoch: ${epoch}, user: ${i}`)
```
vimwitch, line 662: stray comment

packages/contracts/test/verifier.test.ts
```diff
@@ -0,0 +1,558 @@
...
+            _publicSignals[0] = BigInt(publicSignals[0].toString()) + BigInt(1)
+            const isValid = await unirepContract.verifyEpochKeyProof(
+                _publicSignals,
+                proof
+            )
+            expect(isValid).to.be.false
+        }
+    })
+
+    it('should fail to verify an epoch key proof with state tree root', async () => {
```
vimwitch, line 244: this name is incorrect i think, should be "with bad state tree root"

vimwitch (changes requested):
Lots of good tests

packages/contracts/test/attestation.test.ts
```diff
@@ -156,4 +158,791 @@ describe('Attestations', function () {
...
+            expect(_index.toNumber()).to.equal(i)
+            const count = await unirepContract.attesterHashchainTotalCount(
+                attester.address,
+                epoch
+            )
+            expect(count).to.equal(i + 1)
+        }
+    })
+
+    it('should build correct hash chain with the same epoch key', async () => {
```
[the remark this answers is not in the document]
vimwitch: Also I think we should evaluate this based on codepaths. If we're testing a codebpath here that we're not hitting elsewhere then I'm all for it.

packages/contracts/test/userSignup.test.ts
```diff
@@ -76,84 +59,103 @@ describe('User Signup', function () {
         ).to.be.revertedWithCustomError(unirepContract, 'InvalidProof')
     })
 
-    it('sign up should succeed', async () => {
-        const id = new ZkIdentity()
+    it('sign up many users should succeed', async () => {
```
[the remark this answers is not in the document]
vivianjeng: It signs up 3 users per epoch * 3 epochs

packages/contracts/test/attestation.test.ts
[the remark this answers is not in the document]
vivianjeng: Actually it is for this: https://github.com/Unirep/Unirep/pull/186
but it was fixed
I am not sure if we should keep this

> It would be good to build 5 hashchains each containing 1 attestation, and then verify that the final balance output is correct (this may be in another test).

Yeah we can build another tests for this to check the final balance

packages/contracts/contracts/Unirep.sol
vimwitch: >  I don't know if there is any case that a false proof can execute a valid transaction.
Yea I tend to agree. If an application wants to take this as a boolean they can use a `try`/`catch` statement.

packages/contracts/test/attestation.test.ts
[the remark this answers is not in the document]
vimwitch: Right maybe it's good to keep this as separate test.

packages/contracts/test/attestation.test.ts
```diff
@@ -347,21 +318,24 @@ describe('Attestations', function () {
...
-                    .submitAttestation(
-                        epoch,
-                        epochKey,
-                        posRep,
-                        negRep,
-                        graffiti
-                    )
-                    .then((t) => t.wait())
-            }
+            const receipt = await unirepContract
```
vimwitch, line 321: Actually this entire thing can be de-structured in one line
```js
const { timestamp } = await unirepContract
  .connect(attester)
  .submitAttestation(epoch, epochKey, posRep, negRep, graffiti)
  .then((t) => t.wait())
  .then(({ blockNumber }) => ethers.provider.getBlock(blockNumber))
```

This way we avoid having too many local variables

packages/contracts/test/attestation.test.ts
```diff
@@ -1,42 +1,95 @@
...
     EPOCH_TREE_DEPTH,
-    STATE_TREE_DEPTH,
-    NUM_EPOCH_KEY_NONCE_PER_EPOCH,
+    defaultEpochTreeLeaf,
 } from '@unirep/circuits'
+import { defaultProver } from '@unirep/circuits/provers/defaultProver'
 
-import { EPOCH_LENGTH } from '../src'
-import { deployUnirep } from '../deploy'
+function genCircuitInputs(
```
vimwitch, line 23: can this be renamed to something more descriptive? It's unclear what circuit the inputs are for.

vivianjeng: [80c4fc2](https://github.com/Unirep/Unirep/pull/187/commits/80c4fc2fe5a6a850f319ab794d13a070b3129531)

vimwitch (approved):

PR #216, opened 2022-11-20
feat(circuits): circuit options using binary

### LINKED ISSUE #181: Prove many values in 1 field
In the reputation proof and epoch proof there are lots of small public signals like `nonce`, `epoch`, `minRep`, and `proveGraffiti`. These signals can be combined into one signal by specifying ranges of bits that are dedicated to each value.

In a single 256 bit value we can reserve specific numbers of bits for each of these values to have fewer public signals (which will decrease the verification cost). One structure might be like this:

```
[64 bits epoch][8 bits nonce][128 bits minRep][8 bits proveGraffiti][52 bits zero]
```

[Here](https://github.com/vimwitch/zerocompress/blob/main/contracts/Decompress.sol#L22) I have a compression system that relied on bit operations. We can use code like this in the unirep contract to extract single binary values from individual fields. This becomes more useful as we make larger proofs around multiple epoch keys.

We can combine signals in other proofs as well if we define the `epoch` value to be 64 bit (we can also represent it using `Number` in javascript). This means that `epoch` and `attesterId` can be combined into a single signal (64 bits + 160 bits = 224 bits).

### LINKED ISSUE #183: Optionally reveal nonce value
It would be nice to force a user to only use a specific nonce value for proofs. e.g. to make sure the user is only using a single identity for executing operations.

To do this we can add a public signal `nonce` to the proofs. If the signal is zero nothing is being proved, if it is non-zero then the proof was executed using the supplied nonce.

To do this we'll need to change nonces to be 1 indexed instead of 0 indexed. e.g. 3 nonces would be 1, 2, and 3. Not 0, 1, and 2.

Alternatively we could use a `proveNonce` field to indicate whether the nonce was publicly proven.

These small values can be encoded into a single public signal, see #181.

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Closes #181 
Closes #183 

## What is the new behavior?

This PR changes the public signal structure to encode many signals in a single signal. This is done by adding a constant number of constraints. In this PR the number of public signals in the epoch key proof is reduced by 2 while only adding ~10 constraints (a 0.3% increase). The reputation proof gains new functionality (prove max rep, prove 0 rep) while reducing the number of public signals by 3.

Both proofs now allow the epoch key nonce to optionally be revealed.

The proof classes have been moved from `@unirep/contracts` to `@unirep/circuits` so they can be used in the circuits tests.

- [x] Yes
- [ ] No


vimwitch:
@vivianjeng This is ready for review. Needs a few more tests but I think the strategy works. I moved the proof classes into the `@unirep/circuits` package so we can use them in the circuits tests.

vimwitch:
Also needs updated docs

packages/circuits/test/proveReputation.test.ts
```diff
@@ -11,32 +11,206 @@ describe('Prove reputation from attester circuit', function () {
...
+            attesterId,
+            startBalance: { posRep: 10, negRep: 5 },
+            proveZeroRep: 0,
+        })
+        try {
+            await genProofAndVerify(Circuit.proveReputation, circuitInputs)
+            expect(false).to.be.true
+        } catch (err) {}
+    })
+
```
vivianjeng, line 195: In this test it should be `proveZeroRep: 1` and the circuit should throw error
I ran it locally and actually the prove if valid
so I found that `expect(false).to.be.true` is caught in `try-catch`
```
AssertionError: expected false to be true
    at Context.<anonymous> (/Users/zhengyawen/Documents/GitHub/Unirep/packages/circuits/test/proveReputation.test.ts:192:32) {
  showDiff: true,
  actual: false,
  expected: true,
  operator: 'strictEqual'
}
```

In summary, the proof is valid (but we hope it fail) but it doesn't throw error in the test scripts.
So we should re-write all the circuit tests, for example:
```ts
        let error
        try {
            await genProofAndVerify(Circuit.proveReputation, circuitInputs)
        } catch (err) {
            error = err
        }
        expect(error).not.equal(undefined)
```

vivianjeng (changes requested):
sorry for the late response
I think it is good solution to optionally prove many things in one proof
but I found an issue in our circuit test scripts
please check

packages/circuits/test/proveReputation.test.ts
vimwitch: [da8e099](https://github.com/Unirep/Unirep/pull/216/commits/da8e099a5dcd1471a42672cf881a95b2065a6e4d)

packages/circuits/test/proveReputation.test.ts
```diff
@@ -11,32 +11,206 @@ describe('Prove reputation from attester circuit', function () {
...
+        expect(data.nonce.toString()).to.equal('0')
+        expect(data.revealNonce.toString()).to.equal('0')
+        expect(data.attesterId.toString()).to.equal('1')
+        expect(data.proveMinRep.toString()).to.equal('0')
+        expect(data.proveMaxRep.toString()).to.equal('0')
+        expect(data.proveZeroRep.toString()).to.equal('0')
+        expect(data.minRep.toString()).to.equal('0')
+        expect(data.maxRep.toString()).to.equal('0')
+        expect(data.proveGraffiti.toString()).to.equal('0')
+        expect(data.graffitiPreImage.toString()).to.equal('0')
```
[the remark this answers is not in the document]
vimwitch: [99a9961](https://github.com/Unirep/Unirep/pull/216/commits/99a9961983d07cf9f04c1ceecd98d1fa8ccd93ea)

packages/circuits/test/proveReputation.test.ts
```diff
@@ -1,42 +1,233 @@
...
         expect(isValid).to.be.true
+        const data = new ReputationProof(publicSignals, proof)
+        expect(data.epoch.toString()).to.equal(epoch.toString())
+        expect(data.nonce.toString()).to.equal('0')
+        expect(data.revealNonce.toString()).to.equal('0')
+        expect(data.attesterId.toString()).to.equal(attesterId.toString())
+        expect(data.proveMinRep.toString()).to.equal('1')
+        expect(data.proveMaxRep.toString()).to.equal('0')
+        expect(data.proveZeroRep.toString()).to.equal('0')
+        expect(data.minRep.toString()).to.equal('2')
```
vivianjeng, line 67: `minRep` as well

packages/contracts/test/verifier.test.ts
```diff
@@ -118,6 +119,74 @@ describe('Epoch key proof verifier', function () {
...
+            )
+
+            const v = await defaultProver.verifyProof(
+                Circuit.verifyEpochKey,
+                r.publicSignals,
+                r.proof
+            )
+            expect(v).to.be.true
+            const proof = new EpochKeyProof(r.publicSignals, r.proof)
+            const signals = await unirepContract.decodeEpochKeySignals(
```
vivianjeng, line 169: should also test `decodeReputationSignals`

packages/circuits/circuits/proveReputation.circom
```diff
@@ -15,8 +15,8 @@ include "./incrementalMerkleTree.circom";
 include "./modulo.circom";
 
 template ProveReputation(STATE_TREE_DEPTH, EPOCH_TREE_DEPTH, EPOCH_TREE_ARITY, EPOCH_KEY_NONCE_PER_EPOCH, MAX_REPUTATION_SCORE_BITS) {
-    signal input epoch;
-    signal input nonce;
+    // signal input epoch;
+    // signal input nonce;
```
vivianjeng, line 19: before merge to main, we can delete the comment out lines

vivianjeng (changes requested):
A few minor changes please check.
After these changes are made it can be merged to `main`

packages/circuits/test/proveReputation.test.ts
```diff
@@ -1,42 +1,233 @@
...
+        const data = new ReputationProof(publicSignals, proof)
+        expect(data.epoch.toString()).to.equal(epoch.toString())
+        expect(data.nonce.toString()).to.equal('0')
+        expect(data.revealNonce.toString()).to.equal('0')
+        expect(data.attesterId.toString()).to.equal(attesterId.toString())
+        expect(data.proveMinRep.toString()).to.equal('0')
+        expect(data.proveMaxRep.toString()).to.equal('1')
+        expect(data.proveZeroRep.toString()).to.equal('0')
+        expect(data.minRep.toString()).to.equal('0')
+        expect(data.maxRep.toString()).to.equal('4')
```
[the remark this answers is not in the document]
vimwitch: I don't understand what this means?

packages/circuits/circuits/proveReputation.circom
vimwitch: [3433f43](https://github.com/Unirep/Unirep/pull/216/commits/3433f433f990277ebf199e7561afd5bdfeeb958c)

packages/contracts/test/verifier.test.ts
vimwitch: [a585690](https://github.com/Unirep/Unirep/pull/216/commits/a5856906cb6aea95f3349e93adbafcaa32f415c1)

vimwitch:
@vivianjeng I don't understand the other comments, could you explain more?

vivianjeng:
> @vivianjeng I don't understand the other comments, could you explain more?

I mean the same as
```ts
const epoch = 1
const nonce = 2
...
const circuitInputs = genReputationCircuitInput({
    id,
    epoch,
    nonce,
    ...
})
...
expect(data.epoch.toString()).to.equal(epoch.toString())
```

we can set `const proveMinRep = 1` (minRep, proveMaxRep, maxRep... as well) and that verify it with `expect(data.proveMinRep.toString()).to.equal(proveMinRep.toString())`

or we use a 
```ts
const control = {
   epoch: 1,
   nonce: 2,
   proveMinRep: 1,
   ...
}
// if no data is set in `control` and it will be set `0`
```
and use a function to verify if `control` and `data` matches
```ts
function verifyData(control, data) {
   expect(data.epoch.toString()).to.equal(control.epoch.toString())
   ...
}
```

vimwitch:
[5d6be7e](https://github.com/Unirep/Unirep/pull/216/commits/5d6be7e551dfea0df34803a97f02866e47ebeeff)

Sure, these are mostly style changes in the tests. I try to avoid leaving too many style comments unless it's _very_ hard to understand the control flow, or introduces bugs.

vivianjeng (approved):
LGTM

PR #230, opened 2022-11-26
Change `getRepByAttester` arguments

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## What is the new behavior?

Don't accept an `attesterId` in the `getRepByAttester` function. Instead use the attester id specified in the class.

- [x] Yes
- [ ] No


packages/docs/docs/core-api/user-state.md
```diff
@@ -67,10 +67,10 @@ state.getEpochKeys(epoch: number): bigint[]
 
 ## getRepByAttester
 
-Get the reputation balance for an attester in an epoch. By default we use the currently specified attester and the latest epoch.
+Get the reputation balance for the current attester up to an epoch. By default we use the epoch before the current epoch. Reputation received in the current epoch isn't considered final until the user executes a user state transition out of the epoch.
```
vivianjeng, line 70: `epoch before the current epoch` sounds like `epoch = currentEpoch - 1`, is it correct?
Or we can be more specific: `By default we use the latest user transition epoch before the current epoch`

vivianjeng (changes requested):

packages/docs/docs/contracts-api/reputation-proof.md
```diff
@@ -0,0 +1,65 @@
...
+this.proveGraffiti
+```
+
+## graffitiPreImage
+
+The graffiti pre-image controlled by the user. This value is only checked if `proveGraffiti` is truthy.
+
+```ts
+this.graffitiPreImage
+```
```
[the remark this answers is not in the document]
vimwitch: Yes the files need to be moved as well from contracts to circuits.

packages/docs/docs/core-api/user-state.md
vimwitch: > epoch before the current epoch sounds like epoch = currentEpoch - 1, is it correct?
Yes

> By default we use the latest user transition epoch before the current epoch

This isn't correct, if the current epoch is 5, and the user last transitioned to epoch 1, we'll show the reputation up until epoch 4.

vimwitch: The reason we do this is because some other internal functions need to get attestations from `currentEpoch - 1` to do things like epoch transition. I think it might be better to by default show the balance up to `currentEpoch`, then in the internal functions explicitly specify `currentEpoch - 1`

vivianjeng: > This isn't correct, if the current epoch is 5, and the user last transitioned to epoch 1, we'll show the reputation up until epoch 4.

I think it takes the `latestTransitionedEpoch`, and it finds the `epkNullifier` from db
so if the user doesn't perform UST, the nullifier cannot be found

https://github.com/Unirep/Unirep/blob/68c293ce216e15e73395f30b80984d0cc1335775/packages/core/src/UserState.ts#L72

and I think it doesn't make sense if user doesn't perform UST but receive reputation 😧
that means other users/attesters can predict epoch keys and give reputation

vivianjeng: But I think in this function `getRepByAttester` it is fine to query from `currentEpoch - 1`
and this line should be changed
https://github.com/Unirep/Unirep/blob/68c293ce216e15e73395f30b80984d0cc1335775/packages/core/src/UserState.ts#L191

vimwitch: Right `latestTransitionedEpoch` is the default value it processes up to. And we use a `<` operator instead of `<=` so if the `currentEpoch` is passed to the function we process up until `currentEpoch - 1`. We just need to change the default value and check where else it's used and change `<` to `<=`. Then if the user passes `currentEpoch` they'll get the balance up to and including the current epoch.

packages/docs/docs/circuits-api/reputation-proof.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+## control
+
+The control field used for the proof. This field contains many signals binary encoded into an array of 256 bit values. This value is automatically decoded into the other properties on this class.
+
+```ts
+this.control[0]
+this.control[1]
+```
```
vivianjeng, line 124: It will be better to link to the control definition in `circuits#prove-reputation-proof`

packages/core/src/UserState.ts
```diff
@@ -173,22 +170,20 @@ export default class UserState extends Synchronizer {
...
         let negRep = BigInt(0)
         let graffiti = BigInt(0)
         let timestamp = BigInt(0)
-        const attesterId = _attesterId ?? this.attesterId
         const orClauses = [] as any[]
-        const latestTransitionedEpoch = await this.latestTransitionedEpoch()
-        for (let x = 0; x < (toEpoch ?? latestTransitionedEpoch); x++) {
+        const attesterId = this.attesterId
+        const toEpoch = _toEpoch ?? this.calcCurrentEpoch()
+        for (let x = 0; x <= toEpoch; x++) {
```
vivianjeng, line 186: ah actually I prefer
```ts
const toEpoch = _toEpoch ?? this.calcCurrentEpoch() - 1
for (let x = 0; x <= toEpoch; x++) {
```

since it is what user already receive and it won't confuse users when they prove reputation

vimwitch: Ah no the user can only prove up to `latestTransitionedEpoch - 1`, not `currentEpoch - 1`. Maybe it's better to add a helper function like `getProvableRep()`.

vivianjeng: I think user can prove up to `latestTransitionedEpoch` and `latestTransitionedEpoch` always less than `currentEpoch`
so it can be
```ts
const toEpoch = _toEpoch ?? await this.latestTransitionedEpoch()
for (let x = 0; x <= toEpoch; x++) {
```

since `latestTransitionedEpoch()` finds the latest nullifier from `currentEpoch`
but the nullifier `hash(id, attester, currentEpoch)` cannot be existed until next epoch
because the circuit computes nullifiers with `fromEpoch`

https://github.com/Unirep/Unirep/blob/a671ac3dc987ba69c34c6c22bd87294addb49ee0/packages/circuits/circuits/userStateTransition.circom#L152

and the `fromEpoch` cannot equal to `currentEpoch`

vivianjeng: so my point is that
- If it uses `latestTransitionedEpoch` here, the document cannot be `currentEpoch - 1` (Though it is changed. And whatever we use as default `toEpoch` here the document should be clear)
- And the `currentEpoch` rep had better not to be included because it is not provable
- `latestTransitionedEpoch` < `currentEpoch`, so I think it can be the default `toEpoch`

vimwitch: Right that's how it originally was. It's not useful to the developer though because they might want to display the most up to date balance. That's why i suggested adding a different function that is `getProvableRep`.

vimwitch: @vivianjeng planning to leave this as is. We can add a `getProvableRep` function in a separate pr.

vivianjeng: I think we change the `<` to `<=` is useful enough for developer to use `getRepByAttester(currentEpoch)`
and then we don't need another function `getProvableRep`
I think `currentEpoch` as default value might be dangerous(!?) and users might not notice there is another function `getProvableRep`

vimwitch: That would require the dev to do `getRepByAttester(currentEpoch + 1)` to get rep to the current epoch. This is even less clear than having a separate function.

> I think currentEpoch as default value might be dangerous(!?)

I think it's unfair to prospective developers to assume this. If a developer uses a bad value for `posRep`/`negRep` the proof will be invalid, or the tree inclusion will be invalid.

vivianjeng: > That would require the dev to do `getRepByAttester(currentEpoch + 1)` to get rep to the current epoch. This is even less clear than having a separate function.

No we just need `getRepByAttester(currentEpoch)` we can get rep in current epoch
if we change `<` to `<=` (like the current commitment)
https://github.com/Unirep/Unirep/blob/b217d74dd5d56a32a1d1a09e72b99fc2cc84a09d/packages/core/src/UserState.ts#L186

vimwitch: oh right i misread and thought you wanted to change `<=` to `<`.

vimwitch: so the current implementation is fine then?

vivianjeng: I prefer
```ts
const toEpoch = _toEpoch ?? await this.latestTransitionedEpoch()
for (let x = 0; x <= toEpoch; x++) {
```

and use `getRepByAttester(currentEpoch)` to get the updated rep

vivianjeng: but what is the `getRepByAttester(currentEpoch)` used for?
since it can neither be provable nor be spent (in the current version)

vimwitch: It shows the latest user reputation balance. It's an important signal about activity in the system.

vivianjeng: another point that not using `currentEpoch` as default value is that
getting the latest rep by `getRepByAttester(currentEpoch)` is trivial but getting provable rep by `getRepByAttester(latestTranstionedEpoch)` or `getRepByAttester(currentEpoch-1)` is not trivial if user is not really familiar with the protocol

packages/core/src/UserState.ts
```diff
@@ -173,22 +170,20 @@ export default class UserState extends Synchronizer {
...
     /**
      * Get the reputation object from a given attester
-     * @param _attesterId The attester ID that the user queries
      * @param toEpoch The latest epoch that the reputation is accumulated
      * @returns The reputation object
      */
     public getRepByAttester = async (
-        _attesterId?: bigint | string,
-        toEpoch?: number
+        _toEpoch?: number
```
vivianjeng, line 177: it is `_toEpoch` here

packages/docs/docs/core-api/user-state.md
```diff
@@ -67,10 +67,10 @@ state.getEpochKeys(epoch: number): bigint[]
 
 ## getRepByAttester
 
-Get the reputation balance for an attester in an epoch. By default we use the currently specified attester and the latest epoch.
+Get the reputation balance for the current attester up to and including the provided epoch. By default reputation up to and including the current epoch is returned.
 
 ```ts
-state.getRepByAttester(attesterId?: bigint, epoch: number): Promise<Reputation>
+state.getRepByAttester(epoch?: number): Promise<Reputation>
```
vivianjeng, line 73: but it is `epoch` here

packages/docs/docs/core-api/user-state.md
```diff
@@ -67,10 +67,10 @@ state.getEpochKeys(epoch: number): bigint[]
 
 ## getRepByAttester
 
-Get the reputation balance for an attester in an epoch. By default we use the currently specified attester and the latest epoch.
+Get the reputation balance for the current attester up to and including the provided epoch. By default reputation up to and including the current epoch is returned.
```
vivianjeng, line 70: should note that the Rep is not provable? 😅

packages/docs/docs/core-api/user-state.md
vivianjeng: The input name in the function and document should be the same

vivianjeng (changes requested):

vivianjeng (approved):
LGTM

PR #232, opened 2022-11-27
epoch key only proof

### LINKED ISSUE #159: Epoch key proof without state tree
There are instances where we want to prove control of an epoch key, without proving membership in a state tree. For example, proving control of a historical epoch key may be done without a state tree membership proof.

We only care about state tree membership in the case of non-repudiability. If a user proves an epoch key in the current epoch, but doesn't prove state tree membership, then they don't _have_ to receive reputation sent to them.

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Closes #159 

Depends #216 

- [ ] Yes
- [x] No


packages/circuits/test/epochKeyLite.test.ts
```diff
@@ -0,0 +1,190 @@
...
+                    epoch,
+                    nonce
+                ).toString()
+            )
+            expect(data.control).to.equal(
+                (
+                    (BigInt(1) << BigInt(232)) +
+                    (BigInt(attesterId) << BigInt(72)) +
+                    (BigInt(epoch) << BigInt(8)) +
+                    BigInt(nonce)
```
vivianjeng, line 89: can the shift bits be more readable?
e.g.
```ts
(BigInt(attesterId) << BigInt( EPOCH_BITS + NONCE_BITS )) +
BigInt(epoch) << BigInt(NONCE_BITS))
```
or it can be concatenated
```ts
const data = revealNonce +
             attesterId.padStart(ATTESTER_ID_BITS) + 
             epoch.padStart(EPOCH_BITS) + ...
```

vivianjeng (changes requested):
(haven't finished reviewed yet, but some quick comments)

packages/circuits/test/epochKeyLite.test.ts
vimwitch: Yea using constants defined in the proof class would make this more clear.

vimwitch: `padStart` shouldn't work, it has to be a numerical addition

vimwitch: This is out of scope for this pull request, documented in #238

packages/contracts/test/verifier.test.ts
```diff
@@ -47,6 +48,197 @@ const signupUser = async (id, unirepContract, attesterId, account) => {
...
+        const accounts = await ethers.getSigners()
+        unirepContract = await deployUnirep(accounts[0])
+
+        const attester = accounts[1]
+        await unirepContract
+            .connect(attester)
+            .attesterSignUp(EPOCH_LENGTH)
+            .then((t) => t.wait())
+    })
+
```
vivianjeng, line 65: can use snapshot to revert evm like #243

vimwitch: Ah the verifier tests are mostly stateless. I'm not opposed to adding snapshot/revert logic, but i'd rather not block this pr and #239. Need these for `2.0.0-alpha-3` and zkhack next week.

vivianjeng: I am also good to go without snapshot logic
or after this is merged I can change it in #239

vivianjeng (approved):

packages/contracts/test/verifier.test.ts
vimwitch: Ok, there shouldn't be any conflicts. It should be possible to copy/paste the hooks into this `describe` block.

PR #234, opened 2022-11-29
Error codes in doc

### LINKED ISSUE #174: Easier to use custom errors
We need a way for attester developers to be able to identify custom errors more easily. This may be as simple as creating a table with the custom error hash keyed to the name that the user can reference in the documentation.

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: Closed #174 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


vimwitch:
I don't think this is the approach we should take. If we're going to use string error messages they need to be very short, like 2-3 characters. But I think we should keep the error objects like we currently have.

I think a better approach is make a table in the docs showing the error message hash and what the hash means. Each error message is hash is calculated like `keccak256(MessageName,arg1Type,arg2Type)`.

vivianjeng:
But in hardhat environment the error hash is not shown :(



Another solution is that if we change the deployment method from typechain `UnirepFactory` to hardhat `getContractFactory`

https://github.com/Unirep/Unirep/blob/50d165fb1af0543e3fdc74b2f7c142518bdaebe9/packages/contracts/deploy/deploy.ts#L133

and `getContractFactory` can link libraries

The custom error can be displayed
How do you think? @vimwitch

vimwitch:
We should bump our hardhat version, [2.10.2](https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%402.10.2) adds support for showing the custom error selector.

vimwitch:
We should probably bump to 2.11.x, it adds support for post-merge chains.

vivianjeng:
> We should probably bump to 2.11.x, it adds support for post-merge chains.

I found that they only change this file

https://github.com/NomicFoundation/hardhat/blob/fda6bd9b1519f7b9cd433fb40e355a295043a7e0/packages/hardhat-core/src/internal/hardhat-network/stack-traces/error-inferrer.ts#L524

(commitment: https://github.com/NomicFoundation/hardhat/commit/85bd04ee85611611ff511ab52154bf4c2282b9cb)

but not this

https://github.com/NomicFoundation/hardhat/blob/fda6bd9b1519f7b9cd433fb40e355a295043a7e0/packages/hardhat-core/src/internal/hardhat-network/stack-traces/solidity-errors.ts#L274

We can create a PR for hardhat 😅

vivianjeng:
https://github.com/NomicFoundation/hardhat/pull/3384

Okay I think I can create a table for error selectors in this PR. And wait the hardhat PR to be merged.

vimwitch:
This LGTM but I'd like to hold off on publishing until https://github.com/NomicFoundation/hardhat/pull/3384 is done. Maybe we could include a note saying that hardhat support for this is pending?

vivianjeng:
yeah I think it is okay to pend this PR until it is merged and published

vivianjeng:
and in some places there are `require` instead of `revert ErrorMessage()`
https://github.com/Unirep/Unirep/blob/dd870015b0b45fe881d2fc124993a12a2969e343/packages/contracts/contracts/Unirep.sol#L228
will we change all the `require` to `revert ErrorMessage()?` @vimwitch

vimwitch:
I think we can merge this, I'm seeing this in the logs:

```
    body: '{"jsonrpc":"2.0","id":118,"error":{"code":-32603,"message":"Error: VM Exception while processing transaction: reverted with an unrecognized custom error","data":{"message":"Error: VM Exception while processing transaction: reverted with an unrecognized custom error","data":"0xa225656d0000000000000000000000000000000000000000000000010000000000000002"}}}',
```

This is the error selector with the ABI encoded argument.

vivianjeng:
> I think we can merge this, I'm seeing this in the logs:
> 
> ```
>     body: '{"jsonrpc":"2.0","id":118,"error":{"code":-32603,"message":"Error: VM Exception while processing transaction: reverted with an unrecognized custom error","data":{"message":"Error: VM Exception while processing transaction: reverted with an unrecognized custom error","data":"0xa225656d0000000000000000000000000000000000000000000000010000000000000002"}}}',
> ```
> 
> This is the error selector with the ABI encoded argument.

yeah it is the original behavior. I wrote it in the info block in `errors.md`

vivianjeng:
> Oh, then why do we need the hardhat pull request? That's plenty to debug

ah like this. it doesn't show in hardhat node
https://github.com/Unirep/Unirep/pull/234#issuecomment-1331668362
but I think if users use js/ts all the time it can be shown while js/ts returns error

vimwitch:
Ah right, now i remember. This is good to know.

vimwitch (approved):
I think this is ok to merge. Can you post the script you used to generate the event identifiers?

vivianjeng:
> I think this is ok to merge. Can you post the script you used to generate the event identifiers?

It was like
```ts
const iface = new ethers.utils.Interface(ABI)
for (const error of Object.keys(iface.errors)) {
    console.log(error, iface.getSighash(error))
}
```

PR #240, opened 2022-12-02
Get epoch keys fix

vimwitch:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

## What is the new behavior?

Make `getEpochKeys` synchronous and adjust the arguments to make it more useful.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/UserState.ts
+++ b/packages/core/src/UserState.ts
@@ -172,7 +172,7 @@ export default class UserState extends Synchronizer {
                 this.id.identityNullifier,
                 this.attesterId.toString(),
                 epoch,
-                i,
+                nonce,
                 this.settings.epochTreeArity ** this.settings.epochTreeDepth
             )
         }
--- a/packages/core/src/UserState.ts
+++ b/packages/core/src/UserState.ts
@@ -157,7 +157,25 @@ export default class UserState extends Synchronizer {
         return foundLeaf.index
     }
 
-    async getEpochKeys(epoch: bigint) {
+    getEpochKeys(_epoch?: bigint | number, nonce?: number) {
+        const epoch = _epoch ?? this.calcCurrentEpoch()
+        if (
+            typeof nonce === 'number' &&
+            nonce >= this.settings.numEpochKeyNoncePerEpoch
+        ) {
+            throw new Error(
+                `getEpochKeys nonce ${nonce} exceeds max nonce value ${this.settings.numEpochKeyNoncePerEpoch}`
+            )
+        }
+        if (typeof nonce === 'number') {
+            return genEpochKey(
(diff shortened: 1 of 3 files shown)


packages/core/src/UserState.ts
```diff
@@ -157,7 +157,25 @@ export default class UserState extends Synchronizer {
...
 
-    async getEpochKeys(epoch: bigint) {
+    getEpochKeys(_epoch?: bigint | number, nonce?: number) {
+        const epoch = _epoch ?? this.calcCurrentEpoch()
+        if (
+            typeof nonce === 'number' &&
+            nonce >= this.settings.numEpochKeyNoncePerEpoch
+        ) {
+            throw new Error(
+                `getEpochKeys nonce ${nonce} exceeds max nonce value ${this.settings.numEpochKeyNoncePerEpoch}`
```
vivianjeng, line 167: I think we can throw error with prefix like `Synchronizer`
https://github.com/Unirep/Unirep/blob/36a62ffb788bcc109b48ab598ea3f5f6a90066b9/packages/core/src/Synchronizer.ts#L41
But it can be done in another PR since there are many errors without prefix

packages/core/src/UserState.ts
```diff
@@ -157,7 +157,25 @@ export default class UserState extends Synchronizer {
...
+        }
+        if (typeof nonce === 'number') {
+            return genEpochKey(
+                this.id.identityNullifier,
+                this.attesterId.toString(),
+                epoch,
+                nonce,
+                this.settings.epochTreeArity ** this.settings.epochTreeDepth
+            )
+        }
```
vivianjeng, line 178: I think there might also be use cases users want to query all time epoch with `nonce=0`
like all airdropped epoch keys in the old unirep social
(or in the future since epoch key nonce can be revealed)

vimwitch: So is that a change to this PR or a separate issue?

vivianjeng: I think it can be changed in this PR?

vimwitch: So you mean like return all epoch keys with nonce 0?

vimwitch: That seems like something that should be done outside of our library. e.g.

```js
Array(epoch).fill(null).map((_, i) => userState.getEpochKeys(i, 0))
```

vivianjeng: ah okay
I found that there is no `genEpochKey` in `UserState`
so that make sense to return what `genEpochKey` returns

vivianjeng (approved):

PR #246, opened 2022-12-03
Hashchain events

### LINKED ISSUE #242: Add events for hashchain operations
Having events for hashchain operations will make it easier to build responsive applications.

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Closes #242 

## What is the new behavior?

This is 3 changes in 1 pr.
- The first change makes it so that the aggregate epoch key proof doesn't check the inclusion proof when short circuiting. This makes it easier to generate circuit inputs because dummy values don't need to have inclusion proofs matching the latest root.
- The second change adds a `getLeaf` function to the `SparseMerkleTree`
- Third change adds events and tests for hashchain operations to make it easier to observe state offchain.

- [ ] Yes
- [x] No


packages/contracts/test/hashchain.test.ts
```diff
@@ -0,0 +1,365 @@
...
+        const epochKey = '0x123455'
+        const epoch = await unirepContract.attesterCurrentEpoch(
+            attester.address
+        )
+
+        await unirepContract
+            .connect(attester)
+            .submitAttestation(epoch, epochKey, 1, 0, 0)
+            .then((t) => t.wait())
+        const hashchainIndex = 0
```
vivianjeng, line 135: I think one of the problems is that how attester can get the current `hashchainIndex` that is going to be processed
maybe it can be written in the test
(in current contracts tests and core tests we just use `hashchainIndex` that we assign but not from contract status or events)
or in the `Synchronizer` it can handle `HashchainBuilt` and `HashchainProcessed` events so that we don't have to use `hashchainIndex` as input in `genAggregateEpochKeysProof`
https://github.com/Unirep/Unirep/blob/4b12237a1fb8bec43ca6ff57a224b17617238cec/packages/core/src/Synchronizer.ts#L389

packages/contracts/contracts/Unirep.sol
```diff
@@ -585,6 +591,21 @@ contract Unirep is IUnirep, VerifySignature {
         return attester.epochLength;
     }
 
+    function attesterEpochSealed(uint160 attesterId, uint256 epoch)
```
vivianjeng, line 594: we can use it in
https://github.com/Unirep/Unirep/blob/4b12237a1fb8bec43ca6ff57a224b17617238cec/packages/contracts/contracts/Unirep.sol#L338-L342

packages/contracts/test/hashchain.test.ts
vimwitch: Right now they have to manually call `attesterHashchainIndex` on the unirep contract. Adding better handlers can certainly be done, but I'd prefer to do it in a separate pull request.

packages/contracts/test/hashchain.test.ts
```diff
@@ -0,0 +1,365 @@
...
+                .to.emit(unirepContract, 'HashchainProcessed')
+                .withArgs(epoch, attester.address, true)
+        }
+        {
+            const sealed = await unirepContract.attesterEpochSealed(
+                attester.address,
+                epoch
+            )
+            expect(sealed).to.be.true
+        }
```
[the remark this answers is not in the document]
vivianjeng: okay we can keep it like this

packages/contracts/test/hashchain.test.ts
vivianjeng: an issue opened here: #249 and #248

vivianjeng (approved):

PR #263, opened 2022-12-14
fix(core): epoch key range calculation

### LINKED ISSUE #262: Epoch key calculations are incorrect
[Here](https://github.com/Unirep/Unirep/blob/main/packages/core/src/UserState.ts#L174) we calculate the range over which epoch key hashes with be modded. We don't cast these values to `BigInt` so the resulting value is incorrect if it's above 2^64.

We didn't catch this in tests because the ranges we use are < 2^64.

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Closes #262 

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/core/src/Synchronizer.ts
+++ b/packages/core/src/Synchronizer.ts
@@ -503,7 +503,8 @@ export class Synchronizer extends EventEmitter {
     protected async _checkEpochKeyRange(epochKey: string) {
         if (
             BigInt(epochKey) >=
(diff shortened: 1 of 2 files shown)


vivianjeng (approved):
should we add test for this?

vimwitch:
Sorta, I think we should start testing with larger trees so these bugs are automatically caught

PR #266, opened 2022-12-14
test(contracts): get `hashchainIndex` from contract

### LINKED ISSUE #248: [test] Take `hashchainIndex` from contract
Now we define like `hashchainIndex = 0` in test scripts
but it can be query from smart contract with `attesterHashchainProcessedCount`

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: closes #248 

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/packages/contracts/test/hashchain.test.ts
+++ b/packages/contracts/test/hashchain.test.ts
@@ -137,6 +137,7 @@ describe('Hashchain tests', function () {
                 attester.address,
                 epoch
             )
+        expect(hashchainIndex.toNumber()).to.equal(0)
         await expect(unirepContract.buildHashchain(attester.address, epoch))
             .to.emit(unirepContract, 'HashchainBuilt')
             .withArgs(epoch, attester.address, hashchainIndex)
@@ -203,6 +204,7 @@ describe('Hashchain tests', function () {
                 attester.address,
                 epoch
             )
+        expect(hashchainIndex.toNumber()).to.equal(0)
         await expect(unirepContract.buildHashchain(attester.address, epoch))
             .to.emit(unirepContract, 'HashchainBuilt')
             .withArgs(epoch, attester.address, hashchainIndex)
@@ -282,6 +284,7 @@ describe('Hashchain tests', function () {
                     attester.address,
                     epoch
                 )
+            expect(hashchainIndex.toNumber()).to.equal(0)
             await unirepContract
                 .buildHashchain(attester.address, epoch)
                 .then((t) => t.wait())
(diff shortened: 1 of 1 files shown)


packages/contracts/test/hashchain.test.ts
```diff
@@ -132,7 +132,11 @@ describe('Hashchain tests', function () {
             .connect(attester)
             .submitAttestation(epoch, epochKey, 1, 0, 0)
             .then((t) => t.wait())
-        const hashchainIndex = 0
+        const hashchainIndex =
```
vimwitch, line 135: We need to check this value, it should always be 0.
 
It could be like:
```
const hashchainIndex = await unirepContract.attesterHashchainProcessedCount(
  attester.address,
  epoch
)
expect(hashchainIndex).to.equal(0)
```

But this roughly simplifies to `const hashchainIndex = 0`

vimwitch (changes requested):
I think this isn't good for the test. Hardcoding the index to 0 when we're not testing the index logic seems like a good thing. If we're in a clean blockchain state and we're testing the hashchain processing logic we should break if the hashchain index is not what we expect.

Do we have/can we add a test for the `hashchainIndex` specifically? e.g. make a bunch of hashchains and check the index each time?

packages/contracts/test/hashchain.test.ts
vimwitch: Maybe it's good to explicitly load and check. If we hardcode 0 It's possible we make multiple chains and they are the same, then it false passes.

vivianjeng: I think we should add another test for `attesterHashchainProcessedCount` which computes the right `hashchainIndex`.
but here I prefer to load it from contract (if we are not testing for that)
I am thinking that if we want to copy-paste this test code to other place it would be better to show how to get this data.
like if we want to test other thing using `genEpochKey`
if would be hard to know from `genEpochKey(idNullifier, 0, 0)` where one `0` is `epoch` and it can be queried from contract

vimwitch: > I think we should add another test for attesterHashchainProcessedCount which computes the right hashchainIndex

Yes i agree

> I am thinking that if we want to copy-paste this test code to other place it would be better to show how to get this data.

Ehh, i don't think we should write code to optimize for copy/paste-ability. But it's fine to load from contract as long as we have an assertion like

```ts
expect(hashchainIndex).to.equal(0)
```

vimwitch (approved):

PR #268, opened 2022-12-15
Circomlib submodule

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

Submodule approach to circom dependencies. The idea is to not use javascript level packages for circom at all. Instead we manage at the git level and point to a specific commit at the circomlib public origin.

- [ ] Yes
- [ ] No

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,5 @@
+[submodule "circomlib"]
+	path = packages/circuits/circuits/circomlib
+	url = https://github.com/iden3/circomlib.git
+        shallow = true
+        active = true
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "author": "Unirep Team",
     "repository": "https://github.com/Unirep/Unirep",
     "scripts": {
+        "install": "git submodule init && git submodule update",
         "contracts": "yarn workspace @unirep/contracts run",
         "core": "yarn workspace @unirep/core run",
         "circuits": "yarn workspace @unirep/circuits run",
--- a/.prettierignore
+++ b/.prettierignore
@@ -13,3 +13,5 @@ yarn.lock
 .yarn
 packages/contracts/coverage
 abi/
+
+packages/circuits/circuits/circomlib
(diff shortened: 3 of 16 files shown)


.gitmodules
```diff
@@ -0,0 +1,5 @@
+[submodule "circomlib"]
+	path = packages/circuits/circuits/circomlib
+	url = https://github.com/iden3/circomlib.git
+        shallow = true
+        active = true
```
vivianjeng, line 5: can we set to branch `v2.0.5`?

vivianjeng:
and this can be merged to `v1` as well

.gitmodules
vimwitch: unfortunately no, only branches can be set, not tags/commits. But this won't break already released packages. It will just affect our development environments. I think this is okay because we'll get new updates automatically.

If it does break things in the future we can fork the circomlib repo and maintain our own branch. We could also fork circomlib now into the unirep namespace and point to that

package.json
```diff
@@ -11,6 +11,7 @@
     "author": "Unirep Team",
     "repository": "https://github.com/Unirep/Unirep",
     "scripts": {
+        "install": "git submodule init && git submodule update",
```
r1oga, line 14: Good to know that you could make it work with git submodules!
So if new changes were to be pushed to [circomlib](https://github.com/iden3/circomlib), this script is to fetch updates right?

vimwitch: Sorta, by default when you clone a repo the submodule directories are empty. This script initializes the submodule and checks out the branch specified in the `.gitmodules` file.

The submodule state actually isn't persisted to the repo, it's a temporary setting in the local index. But if any changes are made to the submodule contents (e.g. they push a new commit to the circomlib master branch) then the submodule symlink will be marked as changed in any git diffs (so we won't accidentally silently commit submodule changes).

We'll almost certainly make our own fork of circomlib so we can pull in changes as we want, instead of getting them whenever they commit.

Really sucks that we can't point to a commit/tag. We could manually do it by making a long script that does like:

```sh
cd packages/circuits/circuits/circomlib
git checkout v2.0.5
```

But that seems annoying/complicated

.gitmodules
vivianjeng: yeah we can maintain the circomlib update in our repo. (I think we don't have to for it)

vivianjeng (approved):

PR #277, opened 2023-01-03
Bootstrap users and attestations

### LINKED ISSUE #244: Variable state trees/signups
In #243 we revert between tests. We should add helpers to sign up many users before each test so we're not test state trees with 1 entry every time.

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: closes #244 

## What is the new behavior?

I think we can remain some tests that only test its functionality but not the whole states.
and we can have another integration test to test through more epochs and more random behavior

- [ ] Yes
- [x] No


packages/contracts/test/epoch.test.ts
```diff
@@ -1,19 +1,226 @@
...
+    SignupProof,
+    AGGREGATE_KEY_COUNT,
+    AggregateEpochKeysProof,
 } from '@unirep/circuits'
 
 import { EPOCH_LENGTH } from '../src'
 import { deployUnirep } from '../deploy'
+import { defaultProver } from '@unirep/circuits/provers/defaultProver'
+
+async function bootstrapUsers(attester, epoch, unirepContract) {
```
vimwitch, line 27: What do you think of putting this in the `@unirep/utils` package? It could be used for tests and for testnet/mainnet deployments to generate data. It would also prevent duplicate code in contracts/core.

This test data could be helpful for building the explorer and for load testing the system.

cc @CJ-Rose

vivianjeng: or I think we can maintain a `@unirep/test` package? because I think developers can install `@unirep/test` in `devDependencies` instead of like `@unirep/utils` is one of the `dependencies` in `@unirep/core`

and I think we should also bootstrap user state transitions and many attesters data in many epochs 
so there might be many functions we can build in the package

vimwitch: Yes I think it's a good idea

vivianjeng:
update in new PR #288

PR #281, opened 2023-01-06
test: add random submitted graffiti

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

- we should test if the graffiti is 0 that will not overwrite the old graffiti
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/contracts/test/attestation.test.ts
+++ b/packages/contracts/test/attestation.test.ts
@@ -268,9 +268,9 @@ describe('Attestations', function () {
             epochKeys[epochKey] = true
             const posRep = Math.floor(Math.random() * 10)
             const negRep = Math.floor(Math.random() * 10)
-            const graffiti = genRandomSalt()
+            const graffiti = Math.random() < 0.5 ? genRandomSalt() : BigInt(0)
 
-            const { timestamp } = await unirepContract
+            const { timestamp: contractTimestamp } = await unirepContract
                 .connect(attester)
                 .submitAttestation(epoch, epochKey, posRep, negRep, graffiti)
                 .then((t) => t.wait())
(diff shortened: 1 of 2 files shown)


vimwitch (commented):
I'm not a fan of randomly testing zero/non-zero graffiti values. I'm not sure what the best approach is. The most robust would be having multiple copies of each test, but that's a lot of code.

I think we should do #285 and then build fuzz testing with the data fields. That way we're not setting `posRep`/`negRep`/`graffiti`, it's just a data array. Then we can maybe run a test multiple times with multiple data configs?

vivianjeng:
> The most robust would be having multiple copies of each test, but that's a lot of code.

or in this PR we can just add one more test for a non-zero graffiti and then a zero graffiti 

I think we can have the data file for each event, like
```js
{
  ['userSignUp']: {
    idCommitment: 0x11234..., 
    attester: 0x2345...
  },
  ['submitAttestation']: {
    posRep: 5, 
    negRep: 4, 
    graffiti: 0
  },
  ...
}
```
and we can randomly generate this data or use a fixed one for specific test
and we can check if
```js
{
  ['submitAttestation']: {
    posRep: 5, 
    negRep: 4, 
    graffiti: 0x333...
  },
  ['submitAttestation']: {
    posRep: 5, 
    negRep: 4, 
    graffiti: 0
  },
  ...
}
```
will succeed

vivianjeng:
random graffiti can be bootstrap with `bootstrapAttestations`
https://github.com/Unirep/Unirep/blob/5876506dcd5a6dee501abc6d23bb94b49374633e/packages/test/src/index.ts#L108

PR #295, opened 2023-01-27
feat(circuits): circom 2.1.3

### LINKED ISSUE #289: circom 2.1.3
circom 2.1.3 doesn't support this
```
➤ YN0000: error[T3001]: Non quadratic constraints are not allowed!
➤ YN0000:    ┌─ "/home/circleci/project/packages/circuits/circuits/epochKeyLite.circom":22:5
➤ YN0000:    │
➤ YN0000: 22 │     control \ (2 ** 233) === 0;
➤ YN0000:    │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ found here
➤ YN0000:    │
➤ YN0000:    = call trace:
➤ YN0000:      ->EpochKeyLite
➤ YN0000: 
➤ YN0000: previous errors were found
```

we should update the circom version?

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: close #289 

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/packages/circuits/circuits/epochKeyLite.circom
+++ b/packages/circuits/circuits/epochKeyLite.circom
@@ -1,4 +1,4 @@
-pragma circom 2.1.3;
+pragma circom 2.0.0;
 
 include "./circomlib/circuits/bitify.circom";
 include "./circomlib/circuits/comparators.circom";
--- a/packages/circuits/circuits/epochKeyLite.circom
+++ b/packages/circuits/circuits/epochKeyLite.circom
@@ -48,8 +48,10 @@ template EpochKeyLite(EPOCH_KEY_NONCE_PER_EPOCH) {
         nonce_bits.out[x] === 0;
     }
 
-    signal epknonce_divided <-- nonce \ EPOCH_KEY_NONCE_PER_EPOCH;
-    epknonce_divided === 0;
+    component nonce_lt = LessThan(8);
+    nonce_lt.in[0] <== nonce;
+    nonce_lt.in[1] <== EPOCH_KEY_NONCE_PER_EPOCH;
+    nonce_lt.out === 1;
 
     control <== reveal_nonce * 2**232 + attester_id * 2**72 + epoch * 2**8 + reveal_nonce * nonce;
 
--- a/packages/circuits/circuits/epochKeyLite.circom
+++ b/packages/circuits/circuits/epochKeyLite.circom
@@ -48,10 +48,8 @@ template EpochKeyLite(EPOCH_KEY_NONCE_PER_EPOCH) {
         nonce_bits.out[x] === 0;
     }
 
-    component nonce_range = LessThan(8);
-    nonce_range.in[0] <== nonce;
-    nonce_range.in[1] <== EPOCH_KEY_NONCE_PER_EPOCH;
-    nonce_range.out === 1;
+    signal epknonce_divided <-- nonce \ EPOCH_KEY_NONCE_PER_EPOCH;
+    epknonce_divided === 0;
 
     control <== reveal_nonce * 2**232 + attester_id * 2**72 + epoch * 2**8 + reveal_nonce * nonce;
 
--- a/packages/circuits/circuits/epochKeyLite.circom
+++ b/packages/circuits/circuits/epochKeyLite.circom
@@ -1,4 +1,4 @@
-pragma circom 2.0.0;
+pragma circom 2.1.3;
 
 include "./circomlib/circuits/poseidon.circom";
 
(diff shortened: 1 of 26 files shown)


packages/circuits/circuits/epochKeyLite.circom
```diff
@@ -18,21 +18,27 @@ template EpochKeyLite(EPOCH_KEY_NONCE_PER_EPOCH) {
      * 1 bit reveal nonce
      **/
 
-    control \ (2 ** 233) === 0;
+    signal control_divided <-- control \ (2 ** 233);
```
vimwitch, line 21: I think this approach to checking the range of values doesn't work. I thought we could apply a constraint to division by a constant but this was apparently a bug that they fixed in 2.1.3. This division has to be constrained because if the value is > 2^233 then the check at line 37 can be broken.

Now we need to use a `LessThan` component for each of these checks (and convert each value to bits with `Num2Bits`). We actually need to use `Num2Bits_strict` because the input comes from the user/unconstrained. I think it's worth building this and seeing how many constraints it adds.

Another approach is accepting the `control` signal as `control[233]` e.g. a binary array. Then we can iterate over each value and check that (`control[x] == 0 || control[x] == 1`). Then take the relevant sections and convert them to numbers.

^ This should add the fewest constraints but requires us to convert the number to a binary array in javascript.

vimwitch (changes requested):

vimwitch:
Hey @yuriko627 i rebased this branch onto the latest work. I realized that we can pass regular variables as input signals and then build the control output in the circuit. All we need to do is make sure that the input signals are within a certain range. Check the `epochKeyLite.circom` circuit for an example of how I did this. You'll see that each relevant input variable is converted to bits, then if we want to check that the input is 64 bits we assert that all bits > 64 are equal to 0.

We can use the same strategy in the other circuits. Make sure to wipe your local copy of this branch before you pull to avoid conflicts. If you have local changes lmk and we can figure out how to handle it.

packages/circuits/circuits/proveReputation.circom
```diff
@@ -177,7 +196,8 @@ template ProveReputation(STATE_TREE_DEPTH, EPOCH_KEY_NONCE_PER_EPOCH, MAX_REPUTA
 
     /* 4. Check nonce and output epoch key */
 
-    nonce \ EPOCH_KEY_NONCE_PER_EPOCH === 0;
+    signal epknonce_divided <-- nonce \ EPOCH_KEY_NONCE_PER_EPOCH;
+    epknonce_divided === 0;
```
vimwitch, line 200: This check doesn't work, need to use a `LessThan` component. You can use `LessThan(8)` because we constrain the value of `nonce` to 8 bits above.

packages/circuits/circuits/sparseMerkleTree.circom
```diff
@@ -22,7 +24,12 @@ template SMTRootCalc(HEIGHT, ARITY) {
         // start doing checks
         if (i > 0) {
             // range check
-            path_indices[i] \ ARITY === 0;
+            mod[i] = Modulo();
+            mod[i].divisor <== ARITY;
+            mod[i].dividend <== path_indices[i];
+
+            path_indices_divided[i] <== mod[i].quotient;
```
vimwitch, line 31: Why do we need the `path_indices_divided` variable? Here we need to check that `path_indices[i] < ARITY`.

packages/circuits/src/EpochKeyLiteProof.ts
```diff
@@ -38,12 +38,12 @@ export class EpochKeyLiteProof extends BaseProof {
...
     }
 
-    static buildControlInput({ attesterId, epoch, nonce, revealNonce }: any) {
+    static buildControl({ attesterId, epoch, nonce, revealNonce }: any) {
         let control = BigInt(0)
         control += BigInt(revealNonce ?? 0) << BigInt(232)
         control += BigInt(attesterId) << BigInt(72)
         control += BigInt(epoch) << BigInt(8)
-        control += BigInt(nonce)
+        control += BigInt(nonce) * BigInt(revealNonce ?? 0)
```
vimwitch, line 46: nice repurpose of this function

vimwitch (changes requested):
Looks good, couple of small changes in the circuits.

packages/circuits/circuits/sparseMerkleTree.circom
vivianjeng: Maybe we can remove the sparseMerkleTree circuits?

packages/circuits/circuits/proveReputation.circom
vivianjeng: fix here [7b47f60](https://github.com/Unirep/Unirep/pull/295/commits/7b47f602654856d80bec74ea01c9288b0a973422)
and also I proposed that `proveGraffiti` can be in `control1` so that the `proveReputation` can use `epochKeyLite` template

vimwitch: I think that's fine, just make sure to update the [`decodeReputationSignals`](https://github.com/Unirep/Unirep/blob/main/packages/contracts/contracts/Unirep.sol#L464) function.

packages/circuits/circuits/sparseMerkleTree.circom
vimwitch: Yep that works as well, I don't think they're used anywhere else. We should do a bigger cleanup in a different PR.

vimwitch: Oh, but the `SMTRootCalc` template is used in other components. It's not specific to an SMT, it works for any kind of merkle tree.

vivianjeng:
@vimwitch I think we can use a function `decodeEpochKeyControl` to decode the epoch key control
https://github.com/Unirep/Unirep/blob/466837ae9f98747395f015845e6408eaee5df917/packages/contracts/contracts/Unirep.sol#L475-L478

and another `decodeReputationControl` to decode the reputation control
https://github.com/Unirep/Unirep/blob/466837ae9f98747395f015845e6408eaee5df917/packages/contracts/contracts/Unirep.sol#L480-L485

so that we don't have to test `decodeEpochKeyLiteSignals`, `decodeEpochKeySignals` and `decodeReputationSignals` for the same decode logic (epoch key control).

vimwitch:
Also would like to do this in a separate PR. I would like to get these open PRs merged so we can make smaller PRs to prepare an audit commit

PR #303, opened 2023-02-05
Synchronizer ergonomics

### LINKED ISSUE #308: #282 tests should be restored
https://github.com/Unirep/Unirep/pull/282/commits/a0dfefd47c7f7d2ada50a39be1e6d8bc598c8a5c

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Closes #308 

## What is the new behavior?

This pull request improves the performance/stability of the synchronizer. When the unirep contract is deployed an attester with the 0 address is created. This event is used as the starting point for synchronization. The synchronizer now uses a more simple polling strategy. In the browser unirep applications tended to leak memory and become very slow after a while. I suspect this was a result of the websocket subscription approach. Now the synchronizer loads events in batches of (up to) 1000 blocks.

The circuits package has a new class `CircuitConfig` that can be used to specify a config for unirep circuits. This class has a static member `CircuitConfig.default` that contains the default config.

The `UserState` class no longer inherits from `Synchronizer`. Instead `UserState` has a property `sync` that represents the synchronizer. A `UserState` object can be initialized with a synchronizer that already exists. e.g. `new UserState(synchronizer, identity)`. A single synchronizer can be used to power many user state instances.

The synchronizer class support global sync (e.g. loading events for all attesters).

The synchronizer class has fewer required arguments in the constructor. Now a synchronizer can be made like `new Synchronizer({ prover, provider, unirepAddress })`. By default an in memory database will be used.

- [x] Yes
- [ ] No


vimwitch:
Needs lots of docs updates

packages/core/src/Synchronizer.ts
```diff
@@ -181,101 +204,140 @@ export class Synchronizer extends EventEmitter {
...
 
+    async findStartBlock() {
+        // look for the first attesterSignUp event
+        // no events could be emitted before this
+        const filter = this.unirepContract.filters.AttesterSignedUp(
+            this.attesterId
+        )
+        const events = await this.unirepContract.queryFilter(filter)
+        if (events.length === 0) {
+            throw new Error('failed to fetch genesis event')
```
vivianjeng, line 215: with `Synchronizer` prefix?

packages/contracts/test/epoch.test.ts
```diff
@@ -51,39 +48,6 @@ describe('Epoch', function () {
...
-            )
-            const epochTree = await bootstrapAttestations(
-                attester,
-                prevEpoch.toNumber(),
-                unirepContract,
-                {
-                    epkNum: 3,
-                    attestNum: 5,
-                }
-            )
```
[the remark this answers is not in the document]
vimwitch: I think this is checked in the `epochSealing` tests

packages/core/src/Synchronizer.ts
```diff
@@ -16,11 +22,11 @@ type EventHandlerArgs = {
  * the synchronizer will verify the events and then save the states.
  */
 export class Synchronizer extends EventEmitter {
-    protected _db: DB
+    public _db: DB
     prover: Prover
     provider: any
     unirepContract: ethers.Contract
-    attesterId: bigint
+    private _attesterId: bigint
```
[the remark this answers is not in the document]
vimwitch: Yes, the `attesterId` is optional in the constructor. By default this value is `0`, which will cause the synchronizer to load events for all attesters

packages/core/src/Synchronizer.ts
vimwitch: 753001c

packages/core/src/UserState.ts
```diff
@@ -23,34 +23,55 @@ import { Synchronizer } from './Synchronizer'
...
+                  provider: ethers.providers.Provider
+                  _id?: ZkIdentity // TODO: remove this and only accept as second arg
+              }
+            | Synchronizer,
+        id: ZkIdentity
+    ) {
+        if (config instanceof Synchronizer) {
+            if (!id) {
+                throw new Error(
+                    'id must be supplied as second argument when initialized with a sync'
```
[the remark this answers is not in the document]
vimwitch: 753001c

packages/circuits/src/circuits.ts
```diff
@@ -1,4 +1,8 @@
 import { SnarkProof } from '@unirep/utils'
+import { CircuitConfig } from './CircuitConfig'
+
+export const SNARK_SCALAR_FIELD =
+    '21888242871839275222246405745257275088548364400416034343698204186575808495617'
```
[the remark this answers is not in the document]
vimwitch: dc80cbeb6b690b808bf87c54ac67a1f50e8013e6

packages/test/src/index.ts
```diff
@@ -1,161 +1,121 @@
...
+
+// users
+export async function bootstrapUsers(
+    synchronizer: Synchronizer,
+    userCount = 5
+) {
+    const { unirepContract } = synchronizer.unirepContract
+    // synchronizer should be authed to send transactions
+    for (let x = 0; x < userCount; x++) {
+        const userState = new UserState(synchronizer, new ZkIdentity())
```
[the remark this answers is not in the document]
vimwitch: d6119a5c5ab73ea6dd819b55541341fb7b63dad3

packages/docs/docs/core-api/synchronizer.md
```diff
@@ -43,9 +40,11 @@ synchronizer.start(): Promise<void>
 Stop the synchronizer daemon.
 
 ```ts
-synchronizer.stop(): Promise<void>
+synchronizer.stop(): void
 ```
 
+## poll
+
```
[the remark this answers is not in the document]
vimwitch: 563be44

packages/contracts/test/epoch.test.ts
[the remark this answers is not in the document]
vivianjeng: how about move them all to `epoch.test.ts`?
I think they are `epoch` related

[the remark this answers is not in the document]
vimwitch: What do you think of doing this in a different PR?

packages/circuits/scripts/build.mjs
```diff
@@ -0,0 +1,5 @@
+// check if circom is in the path
+
+// ask the user for config values
+
+// build the circuits
```
vivianjeng, line 5: not finish yet?
Or it is a TODO in another PR?

packages/circuits/test/buildOrderedTree.test.ts
```diff
@@ -1,16 +1,17 @@
 import { expect } from 'chai'
 import { IncrementalMerkleTree, hash1, hash5 } from '@unirep/utils'
+import {} from '../config'
```
vivianjeng, line 3: can be removed?

vivianjeng (commented):
some `SNARK_SCALAR_FIELD` in tests still imported from `@unirep/circuits` but not from `CircuitConfig`
I think it would be better to import from `CircuitConfig`
and we should change the `SNARK_SCALAR_FIELD` in the contract

if we change the field we should use another elliptic curve? will it be another precompiled contract?

packages/circuits/scripts/build.mjs
vimwitch: Ah it was just an idea. Like to have an interactive CLI for building circuits and setting config.

vimwitch:
> some SNARK_SCALAR_FIELD in tests still imported from @unirep/circuits but not from CircuitConfig

Yes, the `index.ts` in the circuit package re-exports it like `export * from './CircuitConfig'`

> if we change the field we should use another elliptic curve? will it be another precompiled contract?

Yes, if we switch to BLS for example we would change the field constant. But for now it's the same as before.

vivianjeng (commented):
https://github.com/Unirep/Unirep/blob/7ca10765283dcaceb647a832c4aae66d073d8299/packages/test/package.json#L10-L12
Actually we only need `@unirep/core`, and `utils` and `circuits` are in `core` dependencies.

I think the PR is almost ready to be merged. What we need to fix in another PR should open issues.

packages/contracts/test/epoch.test.ts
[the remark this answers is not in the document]
vimwitch: I'm not really against this change, but I don't think it's worth opening an issue in the backlog. There are lots of other things to do.

packages/core/src/Synchronizer.ts
```diff
@@ -181,101 +204,144 @@ export class Synchronizer extends EventEmitter {
         }
     }
 
+    async findStartBlock() {
+        // look for the first attesterSignUp event
+        // no events could be emitted before this
+        const filter = this.unirepContract.filters.AttesterSignedUp(
+            this.attesterId
+        )
```
vivianjeng, line 212: I think it doesn't have to be a specific `attesterId`
it can just query `AttesterSignedUp` events and stores new attesterId in db
(and the `latestBlocknumber` can also be stored somewhere so that it won't query through whole blocks.

Or how can the synchronizer work with many attesters? can you write a test scripts for that?

and also I think the `poll` function should be tested.

vimwitch: Actually we need to query this event to load the `startTimestamp` and `epochLength` anyway

vivianjeng: Got it! thank you for explaining.
We can merge this first and then add tests for that.
Though I prefer to test a new feature in the same PR.

packages/core/test/synchronizerAttesters.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    before(async () => {
+        const accounts = await ethers.getSigners()
+        unirepContract = await deployUnirep(accounts[0])
+        for (let x = 0; x < 10; x++) {
+            const attester = accounts[x]
+            await unirepContract
+                .connect(attester)
+                .attesterSignUp(EPOCH_LENGTH)
+                .then((t) => t.wait())
+        }
```
vivianjeng, line 23: I prefer that a synchronizer can start first
then attesters sign up and the event should be caught by the synchronizer.

vimwitch: There's no different between starting the synchronizer before emitting the event or after. In either case the codepath is the same.

vimwitch: I did add a test for `AttesterSignedUp`

packages/core/test/synchronizerAttesters.test.ts
```diff
@@ -14,7 +14,7 @@ describe('Synchronizer watch multiple attesters', function () {
     before(async () => {
         const accounts = await ethers.getSigners()
         unirepContract = await deployUnirep(accounts[0])
-        for (let x = 0; x < 10; x++) {
+        for (let x = 0; x < 5; x++) {
```
vivianjeng, line 17: it can be a `const numAttesters  = 5`

vivianjeng (approved):
I think it is good to be merged 👍🏻

PR #313, opened 2023-02-09
refactor(circuits): refactor big comparison component

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/circuits/circuits/bigComparators.circom
+++ b/packages/circuits/circuits/bigComparators.circom
@@ -37,11 +37,18 @@ template BigLessThan() {
     low_lt.in[0] <== low[0].out;
     low_lt.in[1] <== low[1].out;
 
-    component mux = Mux1();
-    mux.s <== high_lt.out;
+    // if high[0] == high[1] then out = low_lt
+    // else out = high_lt
 
-    mux.c[0] <== low_lt.out;
-    mux.c[1] <== 1;
+    component is_high_eq = IsEqual();
+    is_high_eq.in[0] <== high[0].out;
+    is_high_eq.in[1] <== high[1].out;
+
+    component mux = Mux1();
+    mux.s <== is_high_eq.out;
+
+    mux.c[0] <== high_lt.out;
+    mux.c[1] <== low_lt.out;
 
     out <== mux.out;
 }
--- a/packages/circuits/circuits/bigComparators.circom
+++ b/packages/circuits/circuits/bigComparators.circom
@@ -3,51 +3,47 @@ include "./circomlib/circuits/bitify.circom";
 include "./circomlib/circuits/mux1.circom";
 include "./modulo.circom";
 
-//~~ Copy of template with assert(n <= 252) removed
 //~~ support comparisons of numbers up to the field size
 
 template BigLessThan() {
-    // take the number mod 250
-    // check the quotient and mod
-
     signal input in[2];
     signal output out;
 
-    component mod[2];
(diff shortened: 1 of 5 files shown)


packages/circuits/circuits/bigComparators.circom
```diff
@@ -3,51 +3,47 @@ include "./circomlib/circuits/bitify.circom";
...
+    low_lt.in[0] <== low[0].out;
+    low_lt.in[1] <== low[1].out;
 
     component mux = Mux1();
-    mux.s <== quotient_lt.out;
+    mux.s <== high_lt.out;
 
-    mux.c[0] <== remainder_lt.out;
+    mux.c[0] <== low_lt.out;
     mux.c[1] <== 1;
```
vivianjeng, line 44: This should be wrong.
If `high[0] >= high[1]`, then `high_lt.out = 0`
then we should check if `high[0] == high[1]` and then compare `low[0]` and `low[1]`
or it should return `0` rather than `low_lt.out`

vivianjeng (changes requested):
we should add some tests for this to compare many big numbers

packages/circuits/circuits/bigComparators.circom
vimwitch: Ah you're right, lower order bits should only be checked if `high[0] == high[1]`, otherwise it returns 0. I agree, this needs to be tested around the edge boundaries.

packages/circuits/package.json
```diff
@@ -19,6 +19,7 @@
     ],
     "scripts": {
         "pre": "node scripts/prepare",
+        "buildTest": "TEST=1 yarn build",
```
vivianjeng, line 22: I think it can be executed before `test`?
So the test script become `yarn buildTest && mocha -r ts-node/register test/*.test.ts --exit`

vivianjeng (commented):
good tests 👍🏻

packages/circuits/package.json
vimwitch: Ah, i prefer not. I often test using `yarn test --grep 'test name'`, so i might not want to build all circuits.

vivianjeng: but it would fail if developers clone can run `yarn build` and `yarn test`
we should either skip these tests locally without building circuits or running these tests and building circuits.

vimwitch: Sure, maybe build all circuits by default and remove the test only circuits as a part of the `pre` script

vimwitch: [c7bf309](https://github.com/Unirep/Unirep/pull/313/commits/c7bf3099bb1bd7229622f7353c1223bfd61d0f54)

packages/circuits/scripts/prepare.js
```diff
@@ -31,6 +31,12 @@ try {
     )
 } catch (_) {}
 
+// remove test keys
+const zkPath = path.join(__dirname, '../dist/zksnarkBuild')
+fs.readdirSync(zkPath)
+    .filter((filename) => /^bigComparators/.test(filename))
```
vivianjeng, line 37: it would be better if we set the test circuits in a list if we have many test circuits in the future.
but it would be fine to be in this PR.

vivianjeng (approved):
LGTM

PR #332, opened 2023-02-21
feat(core): add `genSealedEpochProof`

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


packages/core/src/Synchronizer.ts
```diff
@@ -742,6 +777,23 @@ export class Synchronizer extends EventEmitter {
             timestamp,
             blockNumber: event.blockNumber,
         })
+
+        db.upsert('Epoch', {
```
vimwitch, line 781: I'd rather not put the `attest` flag on the epoch object. Upserting is wonky with `Epoch` objects because they have a compound index. The anondb implementation needs a single unique key to determine if an existing document exists. This is why in the `handleEpochEnded` function i use a find, then a create or update call after.

Instead of storing `attest: boolean` on the epoch we can do 

```js
db.findOne('Attestation', { where: { attesterId, epoch }})
```

Or we can keep the `attest` field on the `Epoch` if this upsert is changed to a `find` and then `update` or `create`.

packages/core/src/Synchronizer.ts
```diff
@@ -589,6 +594,36 @@ export class Synchronizer extends EventEmitter {
         return preimages
     }
 
+    async genSealedEpochProof(
```
vimwitch, line 597: We should include a note in the docs that this function shouldn't be used in production. Instead use something like [epoch-sealer](https://github.com/unirep/epoch-sealer).

vimwitch (commented):
Looks good, couple small things

packages/core/src/Synchronizer.ts
vivianjeng: good idea

packages/core/src/Synchronizer.ts
vivianjeng: I think attesters can still run there epoch-sealer so that they don't have to rely on a single epoch-sealer

vimwitch: They don't need to rely on the epoch-sealer that we're running. They can run their own copy of epoch-sealer and have it only seal epochs from their attester.

packages/core/src/Synchronizer.ts
```diff
@@ -589,6 +595,54 @@ export class Synchronizer extends EventEmitter {
...
+            throw new Error(
+                `Synchronizer: no attestation is made in epoch ${unsealedEpoch.number}.`
+            )
+        }
+        const epoch = options.epoch ?? unsealedEpoch.number
+        const preimages =
+            options.preimages ??
+            (await this.genEpochTreePreimages(epoch, attesterId))
+        const { circuitInputs } =
+            BuildOrderedTree.buildInputsForLeaves(preimages)
```
vimwitch, line 637: Make sure to pass `settings.epochTreeArity`, `settings.epochTreeDepth` and `settings.fieldCount` so it doesn't use the default config.

packages/docs/docs/core-api/synchronizer.md
```diff
@@ -176,3 +176,17 @@ Get the number of state tree leaves in a certain epoch.
 ```ts
 synchronizer.numStateTreeLeaves(epoch: number): Promise<number>
 ```
+
+## genSealedEpochProof
+
+Generate the sealed epoch proof. See [`sealEpoch`](../contracts-api/unirep-sol.md#sealepoch)
```
vimwitch, line 182: Maybe add a note that this should only be used for small trees? Using snarkjs will be much slower than rapidsnark.

packages/core/test/sealEpoch.test.ts
```diff
@@ -0,0 +1,127 @@
...
+            await ethers.provider.send('evm_revert', [snapshot])
+        })
+    }
+
+    it('should seal the oldest epoch', async () => {
+        const accounts = await ethers.getSigners()
+        const attester = accounts[1]
+        const attesterId = BigInt(attester.address)
+
+        let firstEpoch = 0
```
vimwitch, line 40: `const` for variables that aren't mutated. Here and line 69

vimwitch (changes requested):

PR #339, opened 2023-02-22
Custom signup

### LINKED ISSUE #253: Allow user to receive reputation on signup
Add `posRep`/`negRep`/`graffiti` as public inputs to the signup proof. Figure out how to receive an attestation on signup.

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Closes #253 

- [ ] Yes
- [x] No


packages/contracts/contracts/Unirep.sol
```diff
@@ -99,47 +99,40 @@ contract Unirep is IUnirep, VerifySignature {
...
-            revert AttesterIdNotMatch(uint160(msg.sender));
-        // Verify the proof
-        if (!signupVerifier.verifyProof(publicSignals, proof))
-            revert InvalidProof();
-
-        uint256 identityCommitment = publicSignals[0];
+     * Use this if your application has custom signup proof logic.
+     * e.g. to insert a non-zero data field in the state tree leaf
+     **/
+    function manualUserSignUp(
```
vivianjeng, line 105: I think it can be public input in `signup.circom`
```js
    signal input data[FIELD_COUNT];
    ....
    for (var x = 0; x < FIELD_COUNT; x++) {
      leaf_hasher.data[x] <== data[x];
    }
```
it can make sure all `stateTreeLeaf` are not a random hash

or we can simply not use `signup` circuit
because the `attesterId`, `epoch` are verified on-chain 
and `leaf` and the `idCommitment` is not verified (like `manualUserSignUp`)

vimwitch: I'm confused, what is the suggestion here?

I thought about making it a public input to the `signup` circuit, but applications might want to calculate the data in a specialized way. Like an application might want to give the user secret data by doing `H(idSecret, data)`.

We do need to figure out how to get the starting data into the `UserState`. Right now if a user manually signs up with a custom state tree leaf the pre-image isn't recorded anywhere.

vivianjeng: Right. but what is the purpose of `signup` proof?
attesters can use `manualUserSignUp` to save gas.
and make sure the users `stateTreeLeaf = hash(data)`

vimwitch: Ah that's interesting. I think the biggest case where we want the signup proof is smart contract attesters. Those attesters _could_ implement the signup verifier in their own contract logic, but I think it's nice to include it in the unirep core contract as a base signup case.

I think having both options is nice, but I agree that we should use the manual case more often.

vimwitch: Maybe in the `manualSignUp` instead of `stateTreeLeaf` we accept `uint[]` that are the state tree pre-image and calculate the state tree leaf in the contract?

vivianjeng: I think it also work in `signup.circom`
The attester smart contract can check `H(idSecret, data)` with their own verifier
and the attester submits the proof with `H(idSecret, data)` is one of the public signals
then the `userSignUp` can verify the `signup` proof
(so no `manualSignUp`)

vimwitch: Would this involve verifying 2 proofs onchain?

vivianjeng: either two proofs or like `H(leaf, data)` (public data) which can be solved by a hash on-chain.

packages/contracts/contracts/Unirep.sol
```diff
@@ -105,9 +105,15 @@ contract Unirep is IUnirep, VerifySignature {
...
-            require(initialData[x] < SNARK_SCALAR_FIELD);
+        require(initialData.length <= fieldCount, 'initdatal');
+        for (uint8 x = 0; x < fieldCount; x++) {
+            require(initialData[x] < SNARK_SCALAR_FIELD, 'fieldrange');
+            if (
+                x >= sumFieldCount &&
+                (x - sumFieldCount) % 2 == (sumFieldCount + 1) % 2
+            ) {
+                require(initialData[x] == 0, 'timedatanz');
+            }
```
vivianjeng, line 116: I think we should not mix use of `require` and `revert` custom errors.

vivianjeng (changes requested):
I think the overload `manualSignUp` is more confused.

I suppose the way we give init data to user is through the `signup` proof
and give them data by assigning data to the `leaf_hasher`
https://github.com/Unirep/Unirep/blob/234ed71893af6ac65bc16927f64eed4af686cb4b/packages/circuits/circuits/signup.circom#L29-L31
and publish the data through `publicSignals` or `initialData`

and I think making
```solidity
function manualUserSignUp(
    uint64 epoch,
    uint256 identityCommitment,
    uint256 stateTreeLeaf
) public
```
public doesn't make sense because the attester can random signup users with any `stateTreeLeaf` that might not match the unirep protocol (e.g. different data field or different data format) that makes unirep not **universal**

So I prefer the solution be one `userSignUp()` on-chain (no manual sign up)
and change the circuit to make `data[FIELD_COUNT]` public
so that the user can prove the data(rep) just after signup in unirep
but not receiving attest in signup epoch and prove in the next epoch

vimwitch:
@vivianjeng The manual signup is not just to store initial data, but also to add arbitrary logic to the signup ZK proof. For example in the zketh chat system a user signs up using [this](https://github.com/Unirep/zketh/blob/main/packages/circuits/circuits/signupWithAddress.circom) proof. It proves control of an ethereum address in addition to doing the signup proof logic. If we don't allow manual signup the attester will have to verify 2 proofs onchain for every signup, and one proof will need to sign the other (to make sure a relay doesn't switch proofs).

> public doesn't make sense because the attester can random signup users with any stateTreeLeaf that might not match the unirep protocol (e.g. different data field or different data format) that makes unirep not universal

This is not true if the attester is a smart contract. If an attester smart contract implements their own signup proof they will have logic like [this](https://github.com/Unirep/zketh/blob/main/packages/contracts/contracts/ZKEth.sol#L111) that verifies a proof and _only_ then calls manual signup, using the outputs from the proof.

The function is technically public, but the attester id for the signup will be `msg.sender`, and must be registered as an attester.

If an attester uses manual signup they are responsible for making sure the `stateTreeLeaf` is well formed. But that seems like an okay requirement, the protocol is designed to be extensible in ZK. We _want_ applications to write their own logic extending the protocol. I'd argue this makes the protocol _more_ universal.

--

I think we can add a flag when an attester signs up that enables manual signup. In the case of an EOA attester they might want to have manual signup disabled (though I still think we should ban EOA attesters).

It could be like `attesterSignUp(address attesterId, uint256 epochLength, bool enableManualUserSignUp)`

If `enableManualUserSignUp` is `false` calls to `manualUserSignUp` from `attesterId` will always revert.

vivianjeng:
okay so what about using
```solidity
function userSignUp(
   uint64 epoch,
   uint256 identityCommitment,
   uint256 stateTreeLeaf
)
```
instead of
```solidity
function userSignUp(uint256[] memory publicSignals, uint256[8] memory proof)
```
(and no `manualSignUp` since they are too duplicated and it can be verified by other attesters)

and we can remove the signup proof so the attester can customize everything

like the `attest` function letting the attester verify epoch key proof and everything else.
and the `userSignUp` checks the `msg.sender` like the `attest` function

vimwitch:
I think that is too customizable. Every application would have to write their own signup proof, and deploy their own verifier, even if they don't need any extra logic.

vivianjeng:
> Every application would have to write their own signup proof, and deploy their own verifier, even if they don't need any extra logic.

the unirep contract can still provide `signupVerifier`
and it can be like 
https://github.com/Unirep/Unirep/blob/3622a9ac3ef766fef8c29455041559e259937395/packages/contracts/contracts/Unirep.sol#L582-L585
it can be called by external contracts

but I still think we can provide init data as public inputs in `signup.circom`
and by default all data are `0`
if attester need to verify init data it can be done in attester contracts

vimwitch:
I think that makes sense, but the function would need to look like

```solidity
function userSignUp(
   uint64 epoch,
   uint256 identityCommitment,
   uint256 stateTreeLeaf,
   uint256[] memory data
)
```

The data field values would have to be passed so they could be emitted as attestations to the identity commitment.

Adding custom data to the signup proof requires that we add `fieldCount` public signals to the proof (and thus as calldata). I think maybe we could have a basic signup with 0 data (like the current signup), then a custom signup with data as public signals.

This ends up being similar to the current implementation though. This signup proof with custom data would be good to put in circuitlib.

This all seems out of scope of this PR though.

vivianjeng:
yeah I think these three signup functions can be this one 
```solidity
function userSignUp(
   uint64 epoch,
   uint256 identityCommitment,
   uint256 stateTreeLeaf,
   uint256[] memory data
)
```
so it can handle the most cases.
can it be a zero length array of `data` so that means the 0 data signup?

vimwitch:
>  can it be a zero length array of data so that means the 0 data signup?

Yes, this is how the current `manualSignUp` is implemented

vivianjeng:
so I think 
https://github.com/Unirep/Unirep/blob/3622a9ac3ef766fef8c29455041559e259937395/packages/contracts/contracts/Unirep.sol#L132-L136
and
https://github.com/Unirep/Unirep/blob/3622a9ac3ef766fef8c29455041559e259937395/packages/contracts/contracts/Unirep.sol#L177
can be merged in to one

```solidity
function userSignUp(
   uint64 epoch,
   uint256 identityCommitment,
   uint256 stateTreeLeaf,
   uint256[] memory data
)
```

vimwitch:
This is a larger change than I'd like to include in this PR. It's also a breaking change because it changes the function signature of `userSignUp`. I think this approach might work, but we should consider/build it in another issue/pr.

I still think we should include a default 0 signup function that uses a builtin proof for applications that don't need initial user data. I also think we should be careful with overloading because ethers (and maybe other clients) don't handle this well. Instead of using `contract.userSignUp` a developer must use `contract['userSignUp(uint256,uint256,uint256[])']`.

vivianjeng:
> This is a larger change than I'd like to include in this PR.

I would rather not to include this PR since there is also a overloading of `manualUserSignUp`
and there is no tests for `['manualUserSignUp(uint64,uint256,uint256)']`

>  It's also a breaking change because it changes the function signature of userSignUp

I think it should be fine if we want to include this feature in the future. 

I prefer that we open a new PR to do custom signup and discard this PR. 
and the **best option** for me is
add a data field in signup proof
```js
signal input data[FIELD_COUNT];
```
and use these two functions onchain
```solidity
function userSignUp(
   uint64 epoch,
   uint256 identityCommitment,
   uint256 stateTreeLeaf,
   uint256[] memory data
) public
// and remove userSignUp(publicSignals, proof)
...

function verifySignUpProof(
   uint256[] publicSignals,
   uint256[8] proof
) public
```

And the **second option** for me is no change in circom
and in smart contract
```solidity
function manualUserSignUp(
   uint64 epoch,
   uint256 identityCommitment,
   uint256 stateTreeLeaf,
   uint256[] memory data
) public
// keep original userSignUp

function verifySignUpProof(
   uint256[] publicSignals,
   uint256[8] proof
) public
```

how do you think?

vivianjeng:
and I think the one in this PR
```solidity
function manualUserSignUp(
   uint64 epoch,
   uint256 identityCommitment,
   uint256 stateTreeLeaf,
) public
```
might cause problem if attesters use this to signup users and giving data value
but the synchronizer will not know what data is given to the user
the user might not receive it and hard to reproduce the non-default state tree leaf.

I know in zketh you already knew what you are doing but in other cases or developers who are not familiar with this protocol, making this function public is dangerous.

vimwitch:
> I would rather not to include this PR since there is also a overloading of manualUserSignUp
and there is no tests for `['manualUserSignUp(uint64,uint256,uint256)']`

`['manualUserSignUp(uint64,uint256,uint256)']` absolutely is tested, it's called internally by `userSignUp` and the overloaded version of `manualUserSignUp`. I can remove this overload no problem though.

> how do you think?

What you're proposing is almost exactly what exists in this PR. It's simply removing the existing `userSignUp` that takes a proof, and replacing it with the implementation of `manualSignUp`.

> I know in zketh you already knew what you are doing but in other cases or developers who are not familiar with this protocol, making this function public is dangerous.

zketh uses the overload with `initialData` so that the `UserState` will pick up the data and make it accessible/provable.

I think the problem with this PR is the `manualSignUp` with no initial data is public. I made it private in the latest commit and removed the overload please take a look.

packages/contracts/contracts/Unirep.sol
vivianjeng: @vimwitch please also fix the `require` statements to `revert` custom errors.
thanks

packages/contracts/contracts/Unirep.sol
```diff
@@ -98,6 +98,37 @@ contract Unirep is IUnirep, VerifySignature {
...
+     **/
+    function manualUserSignUp(
+        uint64 epoch,
+        uint256 identityCommitment,
+        uint256 stateTreeLeaf,
+        uint256[] memory initialData
+    ) public {
+        _userSignUp(epoch, identityCommitment, stateTreeLeaf);
+        require(initialData.length <= fieldCount, 'initdatal');
+        for (uint8 x = 0; x < fieldCount; x++) {
```
vivianjeng, line 113: can it be `initialData.length`?
The length of `initialData` might be variable?

packages/contracts/test/userSignup.test.ts
```diff
@@ -307,4 +335,144 @@ describe('User Signup', function () {
...
+
+        const leaf = genStateTreeLeaf(
+            id.secretHash,
+            attester.address,
+            contractEpoch.toNumber(),
+            data
+        )
+        const tx = await unirepContract
+            .connect(attester)
+            ['manualUserSignUp(uint64,uint256,uint256,uint256[])'](
```
vivianjeng, line 370: it can use only `manualUserSignUp(...)`

packages/contracts/test/userSignup.test.ts
```diff
@@ -307,4 +335,144 @@ describe('User Signup', function () {
...
+                        (config.sumFieldCount + 1) % 2
+                ) {
+                    return 1
+                }
+                return i + 100
+            })
+
+        const tx = unirepContract
+            .connect(attester)
+            ['manualUserSignUp(uint64,uint256,uint256,uint256[])'](
```
vivianjeng, line 429: same here

packages/contracts/test/userSignup.test.ts
```diff
@@ -307,4 +335,144 @@ describe('User Signup', function () {
...
+        const config = await unirepContract.config()
+
+        const id = new ZkIdentity()
+        const contractEpoch = await unirepContract.attesterCurrentEpoch(
+            attester.address
+        )
+
+        const tx = unirepContract
+            .connect(attester)
+            ['manualUserSignUp(uint64,uint256,uint256,uint256[])'](
```
vivianjeng, line 450: same here

packages/contracts/test/userSignup.test.ts
```diff
@@ -307,4 +335,144 @@ describe('User Signup', function () {
...
+        const attester = accounts[1]
+
+        const id = new ZkIdentity()
+        const contractEpoch = await unirepContract.attesterCurrentEpoch(
+            attester.address
+        )
+
+        const tx = unirepContract
+            .connect(attester)
+            ['manualUserSignUp(uint64,uint256,uint256,uint256[])'](
```
vivianjeng, line 470: same here

packages/core/test/userSignup.test.ts
```diff
@@ -100,4 +100,57 @@ describe('User Signup', function () {
...
+        const leaf = genStateTreeLeaf(
+            id.secretHash,
+            attester.address,
+            contractEpoch.toNumber(),
+            data
+        )
+
+        await unirepContract
+            .connect(attester)
+            ['manualUserSignUp(uint64,uint256,uint256,uint256[])'](
```
vivianjeng, line 142: same here

packages/docs/docs/contracts-api/unirep-sol.md
```diff
@@ -23,6 +23,22 @@ function userSignUp(
...
+:::caution
+`msg.sender` must be the attester.
+:::
+
+```sol
+function manualUserSignUp(
+  uint64 epoch,
+  uint256 identityCommitment,
+  uint256 stateTreeLeaf
+) public
```
vivianjeng, line 39: `uint256[] memory initialData`

vivianjeng (changes requested):

packages/contracts/package.json
```diff
@@ -21,7 +21,7 @@
         "build": "ts-node ./scripts/buildVerifiers.ts && yarn compile && yarn abi && hardhat typechain && tsc",
         "abi": "node ./scripts/abi",
         "compile": "hardhat compile",
-        "test": "DEPLOY_DELAY=0 hardhat test ./test/*.test.ts",
+        "test": "DEPLOY_DELAY=0 hardhat test ./test/userSignup.test.ts",
```
vivianjeng, line 24: should be reverted

vivianjeng (changes requested):

vimwitch:
@vivianjeng this is ready for review

vivianjeng (approved):
LGTM

PR #341, opened 2023-02-23
adding explorer link

ChialiT:
- add link in top nav and footer
- wrote a page for create-unirep-app


ChialiT:
@AnthonyMadia let's add more into the "create-unirep-app"

AnthonyMadia:
For a  "quick start" (and the docs as a whole), we should clarify what assumptions we are making about the developer: assume they know nothing about zk?, something about zk?, etc. This assumption will inform what we include in the docs. Good to put ourselves in the mindset of developer who has never heard of unirep before. 

I'm a fan of Hardhat's [docs](https://hardhat.org/hardhat-runner/docs/getting-started#overview) because of the added context around their repo in their "quick start", a nice overview of things, and then guides after the quickstart to answer more questions about how their software works. 

misc. ideas:
- show how to submit an attestation
- prove a certain piece of data

Showing those two things are probably some of the most important things to demonstrate. More context can be added as well when doing this like showing the various abstractions we have for making proofs in the prover.

ChialiT:
I like your thought @AnthonyMadia let's jam together for this.

PR #345, opened 2023-02-28
fix(core): customizable sync rate

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

Allow synchronizer batch sizes to be customized with a variable.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/Synchronizer.ts
+++ b/packages/core/src/Synchronizer.ts
@@ -37,6 +37,7 @@ export class Synchronizer extends EventEmitter {
 
     private pollId: string | null = null
     public pollRate: number = 5000
+    public blockRate: number = 100000
 
     private setupComplete = false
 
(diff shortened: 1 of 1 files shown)


packages/core/src/Synchronizer.ts
```diff
@@ -37,6 +37,7 @@ export class Synchronizer extends EventEmitter {
 
     private pollId: string | null = null
     public pollRate: number = 5000
+    public blockRate: number = 100000
```
vivianjeng, line 40: how do you think `blockRate` to be an input option of Synchronizer?

vimwitch: I think it's okay, but it should be an optional input. Most developers shouldn't need to change this though, if someone needs to change it they'll probably have a specific reason, and look for this specifically.

To me it's easier if the constructor has fewer arguments, it seems easier to read in the docs.

vivianjeng: but it is in `@unirep/core` and if developers want to use the synchronizer they have to dig into `node_modules` to change the `blockRate`.
Yeah it should be optional
or we can provide another function like `changeBlockRate`  (not in constructor) in synchronizer to set the value

vimwitch: No, they can in their code do like
```js
const synchronizer = new Synchronizer({ ... })
synchronizer.blockRate = 9999999
```

Such a change will take effect immediately. We just need to put the property in the documentation.

vivianjeng (approved):

PR #330, opened 2023-02-19
External nullifier proofs

yuriko627:
## Pull request checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ ] Build (`yarn build`) was run locally and any changes were pushed
- [ ] Lint (`yarn lint --check`) has passed locally and any fixes were made for failures

## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Issue Number: closes https://github.com/Unirep/Unirep/issues/259

## What is the new behavior?

- 
-
-

## Does this introduce a breaking change?

- [ ] Yes
- [ ] No

PR #351, opened 2023-03-07
WIP: add prevent double action circuit

yuriko627:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #330 

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/packages/circuits/circuits/preventDoubleAction.circom
+++ b/packages/circuits/circuits/preventDoubleAction.circom
@@ -1,0 +1,187 @@
+pragma circom 2.0.0;
+
+include "./circomlib/circuits/poseidon.circom";
+include "./incrementalMerkleTree.circom";
+include "./circomlib/circuits/gates.circom";
+include "./circomlib/circuits/poseidon.circom";
+include "./sparseMerkleTree.circom";
+
+template PreventDoubleAction(STATE_TREE_DEPTH, EPOCH_KEY_NONCE_PER_EPOCH, MAX_REPUTATION_SCORE_BITS) {
+    // Global state tree
+    signal input state_tree_indexes[STATE_TREE_DEPTH];
+    signal input state_tree_elements[STATE_TREE_DEPTH][1];
+    signal output state_tree_root;
+
+    // Global state tree leaf: Identity & user state root
+    signal input identity_secret;
+
+    // Optionally reveal nonce, epoch, attester_id
+    signal input control[2];
+    signal output control_output[2];
+    signal output epoch_key;
+
+    signal input pos_rep;
+    signal input neg_rep;
+    signal input graffiti;
+    signal input timestamp;
+    signal input graffiti_pre_image;
+
+    /**
+     * control[0] structure:
+     * 8 bits nonce
+     * 64 bits epoch
+     * 160 bits attester_id
+     * 1 bit reveal nonce
+     * 1 bit prove_graffiti
+     **/
+    control[0] \ (2 ** 234) === 0;
+    signal prove_graffiti <-- (control[0] \ 2 ** 233) & 1;
+    signal reveal_nonce <-- (control[0] \ 2 ** 232) & 1;
+    signal attester_id <-- (control[0] \ 2 ** 72) & (2 ** 160 - 1);
+    signal epoch <-- (control[0] \ 2 ** 8) & (2 ** 64 - 1);
+    signal nonce <-- control[0] & (2 ** 8 - 1);
+
+    // individual range check
+    prove_graffiti \ 2 === 0;
+    reveal_nonce \ 2 === 0;
+    attester_id \ 2**160 === 0;
+    epoch \ 2**64 === 0;
+    nonce \ 2**8 === 0;
+
+    // check extended value
+    control[0] === prove_graffiti * 2 ** 233 + reveal_nonce * 2**232 + attester_id * 2**72 + epoch * 2**8 + nonce;
+
+    /**
+     * control[1] structure:
+     * 64 bits min_rep
+     * 64 bits max_rep
+     * 1 bit prove_min_rep
+     * 1 bit prove_max_rep
+     * 1 bit prove_zero_rep
+     **/
+    control[1] \ (2 ** 131) === 0;
+    signal prove_zero_rep <-- (control[1] \ 2 ** 130) & 1;
+    signal prove_max_rep <-- (control[1] \ 2 ** 129) & 1;
+    signal prove_min_rep <-- (control[1] \ 2 ** 128) & 1;
+    signal max_rep <-- (control[1] \ 2 ** 64) & (2 ** 64 - 1);
+    signal min_rep <-- control[1] & (2 ** 64 - 1);
+
+    // individual range check
+    prove_zero_rep \ 2 === 0;
+    prove_max_rep \ 2 === 0;
+    prove_min_rep \ 2 === 0;
+    max_rep \ 2**64 === 0;
+    min_rep \ 2**64 === 0;
+
+    // check extended value
+    control[1] === prove_zero_rep * 2 ** 130 + prove_max_rep * 2**129 + prove_min_rep * 2**128 + max_rep * 2**64 + min_rep;
+
+    control_output[0] <== prove_graffiti * 2 ** 233 + reveal_nonce * 2**232 + attester_id * 2**72 + epoch * 2**8 + reveal_nonce * nonce;
+    control_output[1] <== control[1];
+
+    /* 1. Check if user exists in the Global State Tree*/
+    // Compute user state tree root
+    component leaf_hasher = Poseidon(7);
+    leaf_hasher.inputs[0] <== identity_secret;
+    leaf_hasher.inputs[1] <== attester_id;
+    leaf_hasher.inputs[2] <== epoch;
+    leaf_hasher.inputs[3] <== pos_rep;
+    leaf_hasher.inputs[4] <== neg_rep;
+    leaf_hasher.inputs[5] <== graffiti;
+    leaf_hasher.inputs[6] <== timestamp;
+
+
+    component merkletree = MerkleTreeInclusionProof(STATE_TREE_DEPTH);
+    merkletree.leaf <== leaf_hasher.out;
+    for (var i = 0; i < STATE_TREE_DEPTH; i++) {
+        merkletree.path_index[i] <== state_tree_indexes[i];
+        merkletree.path_elements[i] <== state_tree_elements[i];
+    }
+    state_tree_root <== merkletree.root;
+
+    /* 2. Check if user has reputation greater than min_rep */
+    // if proving min_rep > 0, check if pos_rep >= neg_rep + min_rep
+    component min_rep_check = GreaterEqThan(MAX_REPUTATION_SCORE_BITS);
+    min_rep_check.in[0] <== pos_rep;
+    min_rep_check.in[1] <== neg_rep + min_rep;
+
+    component if_not_prove_min_rep = IsZero();
+    if_not_prove_min_rep.in <== prove_min_rep;
+
+    component output_rep_check = OR();
+    output_rep_check.a <== if_not_prove_min_rep.out;
+    output_rep_check.b <== min_rep_check.out;
+
+    output_rep_check.out === 1;
+
+    /* 3. Check if user has reputation less than max_rep */
+    // if proving max_rep > 0, check if neg_rep >= pos_rep + min_rep
+    component max_rep_check = GreaterEqThan(MAX_REPUTATION_SCORE_BITS);
+    max_rep_check.in[0] <== neg_rep;
+    max_rep_check.in[1] <== pos_rep + max_rep;
+
+    component if_not_prove_max_rep = IsZero();
+    if_not_prove_max_rep.in <== prove_max_rep;
+
+    component max_rep_check_out = OR();
+    max_rep_check_out.a <== if_not_prove_max_rep.out;
+    max_rep_check_out.b <== max_rep_check.out;
+
+    max_rep_check_out.out === 1;
+
+    /* 4. Check if user has net 0 reputation */
+    component zero_rep_check = IsEqual();
+    zero_rep_check.in[0] <== pos_rep;
+    zero_rep_check.in[1] <== neg_rep;
+
+    component if_not_prove_zero_rep = IsZero();
+    if_not_prove_zero_rep.in <== prove_zero_rep;
+
+    component zero_rep_check_out = OR();
+    zero_rep_check_out.a <== if_not_prove_zero_rep.out;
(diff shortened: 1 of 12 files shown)


packages/circuits/circuits/preventDoubleAction.circom
```diff
@@ -0,0 +1,76 @@
...
+    epoch_key_lite.reveal_nonce <== reveal_nonce;
+    epoch_key_lite.attester_id <== attester_id;
+    epoch_key_lite.epoch <== epoch;
+    epoch_key_lite.nonce <== nonce;
+    epoch_key_lite.sig_data <== sig_data;
+    control <== epoch_key_lite.control;
+    epoch_key <== epoch_key_lite.epoch_key;
+
+    // /* 3. Check identity nullifier and proposal index*/
+    signal input identityNullifier;
```
vimwitch, line 64: You're taking an `identity_secret` and an `identity_nullifier`. You should instead take a nullifier and trapdoor and calculate the secret from that. Otherwise the user could supply a secret/nullifier that don't match.

Also prefer using `snake_case` for variable names in circuits.

packages/circuits/circuits/preventDoubleAction.circom
```diff
@@ -0,0 +1,76 @@
...
+    signal input proposalIndex;
```
vimwitch, line 65: We should try to match the naming that semaphore uses for these components. `externalNullifier` and `identityNullifier` are the values supplied by the user (inputs to the hash). The output is just `nullifier`.

So this variable should be `external_nullifier`.

Also, try to put input signal declarations at the top of the file so it's more clear what inputs the circuit takes.

packages/circuits/src/PreventDoubleActionProof.ts
```diff
@@ -0,0 +1,99 @@
...
+        stateTreeRoot: 1,
+        control0: 2,
+        control1: 3,
+        graffitiPreImage: 4,
+        data: 5,
+    }
+    public epochKey: BigNumberish
+    public stateTreeRoot: BigNumberish
+    public control0: BigNumberish
+    public control1: BigNumberish
```
vimwitch, line 21: I think there is a single `control` output in this circuit.

packages/circuits/src/PreventDoubleActionProof.ts
```diff
@@ -0,0 +1,99 @@
...
+    public epoch: BigNumberish
+    public revealNonce: BigNumberish
+    public nonce: BigNumberish
+    public attesterId: BigNumberish
+    public proveMinRep: BigNumberish
```
vimwitch, line 26: Unused, same for line 27-32

packages/circuits/src/PreventDoubleActionProof.ts
```diff
@@ -0,0 +1,99 @@
...
+import { SnarkProof } from '@unirep/utils'
+import { BigNumberish } from '@ethersproject/bignumber'
+import { BaseProof } from './BaseProof'
+
+/**
+ * The reputation proof structure that helps to query the public signals
+ */
+export class PreventDoubleActionProof extends BaseProof {
+    readonly idx = {
+        epochKey: 0,
```
vimwitch, line 11: This mapping is not correct, it includes outputs for a different circuit and doesn't include outputs for this circuit.

packages/circuits/src/PreventDoubleActionProof.ts
```diff
@@ -0,0 +1,99 @@
...
+        this.revealNonce = (BigInt(this.control0) >> BigInt(232)) & BigInt(1)
+        this.attesterId =
+            (BigInt(this.control0) >> BigInt(72)) &
+            ((BigInt(1) << BigInt(160)) - BigInt(1))
+        this.epoch =
+            (BigInt(this.control0) >> BigInt(8)) &
+            ((BigInt(1) << BigInt(64)) - BigInt(1))
+        this.nonce =
+            BigInt(this.control0) & ((BigInt(1) << BigInt(8)) - BigInt(1))
+        this.minRep =
```
vimwitch, line 59: Variables being decoded here that don't exist in the proof.

packages/circuits/src/PreventDoubleActionProof.ts
```diff
@@ -0,0 +1,99 @@
...
+        this.proveMinRep = (BigInt(this.control1) >> BigInt(128)) & BigInt(1)
+        this.proveMaxRep = (BigInt(this.control1) >> BigInt(129)) & BigInt(1)
+        this.proveZeroRep = (BigInt(this.control1) >> BigInt(130)) & BigInt(1)
+        this.proveGraffiti = (BigInt(this.control1) >> BigInt(131)) & BigInt(1)
+        this.graffitiPreImage = _publicSignals[this.idx.graffitiPreImage]
+        this.data = _publicSignals[this.idx.data]
+        this.circuit = Circuit.proveReputation
+    }
+
+    static buildControl({
```
vimwitch, line 73: Don't redefine the `buildControl` here, you can use the one in `EpochKeyLiteProof.ts`

packages/circuits/test/preventDoubleAction.test.ts
```diff
@@ -0,0 +1,376 @@
+import { expect } from 'chai'
```
vimwitch, line 1: Should test the `PreventDoubleActionProof` class in this file.

vimwitch (changes requested):
@vivianjeng I think we can move this into `circuitlib` once we have a build system there, but for now it can just be in the `@unirep/circuits` package and not the contract.

@yuriko627 Make sure to rebase this branch on the latest `main`.

packages/circuits/circuits/preventDoubleAction.circom
```diff
@@ -0,0 +1,81 @@
...
+    signal output state_tree_root;
+
+    // Optionally reveal nonce, epoch, attester_id
+    signal output control;
+
+    signal input identity_nullifier;
+    signal input external_nullifier;
+    signal output nullifier;
+
+    signal input trapdoor;
```
vimwitch, line 38: Name this consistently with the other e.g. `identity_trapdoor`

packages/circuits/circuits/preventDoubleAction.circom
```diff
@@ -0,0 +1,81 @@
...
+include "./identity.circom";
+
+template PreventDoubleAction(STATE_TREE_DEPTH, EPOCH_KEY_NONCE_PER_EPOCH, FIELD_COUNT, EPK_R) {
+
+    // Global state tree
+    signal input state_tree_indexes[STATE_TREE_DEPTH];
+    signal input state_tree_elements[STATE_TREE_DEPTH];
+
+    // Global state tree leaf: Identity & user state root
+    signal input identity_secret;
```
vimwitch, line 17: Don't accept the identity secret as an input, use the calculated value from the identity nullifier/trapdoor. The `IdentityCommitment` component has a property `secret` that you can use.

packages/circuits/src/PreventDoubleActionProof.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * The prevent double action proof structure that helps to query the public signals
+ */
+export class PreventDoubleActionProof extends BaseProof {
+    readonly idx = {
+        epochKey: 0,
+        stateTreeRoot: 1,
+        control: 2,
+        data: 3,
```
vimwitch, line 14: I think this is ordered incorrectly, `data` should be the last element because it's a public input.

packages/circuits/src/PreventDoubleActionProof.ts
```diff
@@ -0,0 +1,58 @@
...
+    public revealNonce: BigNumberish
+    public attesterId: BigNumberish
+    public epoch: BigNumberish
+    public nonce: BigNumberish
+    public epochKey: BigNumberish
+    public stateTreeRoot: BigNumberish
+    public control: BigNumberish
+    public data: BigNumberish
+    public nullifier: BigNumberish
+    public identity_commitment: BigNumberish
```
vimwitch, line 27: In js we use lower camel case, so this should be `identityCommitment`

packages/circuits/test/preventDoubleAction.test.ts
```diff
@@ -0,0 +1,348 @@
...
+
+            const p = new PreventDoubleActionProof(publicSignals, proof)
+            expect(p.epoch.toString()).to.equal(epoch.toString())
+            expect(p.nonce.toString()).to.equal('0')
+            expect(p.revealNonce.toString()).to.equal('0')
+            expect(p.attesterId.toString()).to.equal(attesterId.toString())
+        }
+    })
+
+    // it('should reveal nonce value', async () => {
```
vimwitch, line 88: Need all of this uncommented.

packages/circuits/test/preventDoubleAction.test.ts
```diff
@@ -0,0 +1,348 @@
...
+    //     for (let nonce = 0; nonce < NUM_EPOCH_KEY_NONCE_PER_EPOCH; nonce++) {
+    //         const attesterId = BigInt(10210)
```
vimwitch, line 90: Also make sure to add tests for `PreventDoubleActionProof`. Like pass in a unique value for each input and check that the variables in the `PreventDoubleActionProof` match what is expected.

packages/utils/src/crypto.ts
```diff
@@ -129,3 +129,18 @@ export const genEpochTreeLeaf = (
...
 }
+
+export const genNullifier = (
+    identityNullifier: bigint,
+    externalNullifier: number | bigint
+): bigint => {
+    return hash2([BigInt(identityNullifier), BigInt(externalNullifier)])
+}
+
+export const genProposalId = (
```
vimwitch, line 140: I think this is named incorrectly, this looks like an identity commitment calculation?

vimwitch (changes requested):
good progress

vimwitch:
Make sure to mark the comments as "resolved" as you go, unless you have questions/comments about it.

packages/circuits/circuits/preventDoubleAction.circom
yuriko627: what would be the appropriate value for `genEpochKey: (identitySecret: bigint, attesterId: bigint | string, epoch: bigint | number, nonce: bigint | number)`'s first argument in the test? would that be `id.secretHash`?

vimwitch: yes

vimwitch: Oh I didn't notice that. That is erroneous, you should be using `identity_nullifier` to calculate the `identity_commitment`. The commitment never changes so the secret should never change. The [semaphore circuit](https://github.com/semaphore-protocol/semaphore/blob/main/packages/circuits/semaphore.circom#L47) is a good reference.

yuriko627: That makes sense! Thank you for the reference.

packages/circuits/test/preventDoubleAction.test.ts
vimwitch: It needs a bit more, it should check all the output variables, like `epochKey` and `identityCommitment` as well as handle the case of `revealNonce` being both `0` and `1`.

We should also consider not forcing the user to reveal an epoch key, but that can be added in later.

yuriko627: so does this look right for checking `epochKey` field?
`expect(p.epochKey.toString()).to.equal(genEpochKey(id.secretHash, attesterId, epoch, nonce).toString())`
Also I looked for a reference, but I didn't find one. Why do we not check public output fields of a proof in other test?

vimwitch: Yes that looks correct.

Ah yes in other tests we seem to only check the control values. That's because someone (me) was not paying attention. Best to check the output mappings in at least 1 place to avoid silently breaking it by rearranging outputs.

packages/utils/src/crypto.ts
```diff
@@ -129,3 +129,18 @@ export const genEpochTreeLeaf = (
     }
     return polysum
 }
+
+export const genNullifier = (
```
vimwitch, line 133: I think we shouldn't add these functions to the utils package. The nullifier can be calculated using just `hash2` and the id commitment can be calculated using `@semaphore-protocol/identity`

vimwitch (changes requested):
Looks good, just needs to take out the functions from utils and rebase

packages/utils/src/crypto.ts
yuriko627: Okay, that makes sense. Am I supposed to rebase this branch to main? `git rebase main` says `Current branch prevent-double-action is up to date.`

vimwitch: Yes, make sure to pull the latest `main` before trying to rebase

yuriko627: ohh sorry for such a dumb question. okay i'll rebase

packages/circuits/package.json
```diff
@@ -37,7 +37,7 @@
         "snarkjs": "^0.5.0"
     },
     "devDependencies": {
-        "@semaphore-protocol/identity": "^3.5.0",
+        "@semaphore-protocol/identity": "^3.6.0-beta.3",
```
vimwitch, line 40: This should be left at `3.5.0`

packages/utils/src/crypto.ts
```diff
@@ -6,6 +6,31 @@ export const SNARK_SCALAR_FIELD =
     '21888242871839275222246405745257275088548364400416034343698204186575808495617'
 export const F = BigInt(SNARK_SCALAR_FIELD)
 
+// Ordered merkle tree polysum constant
```
vimwitch, line 9: Lots of these changes are wrong, the various `R` values should be removed, same with `genUserStateTransitionNullifier`, `genEpochNullifier`.

The implementation of `genStateTreeLeaf` and `genEpochTreeLeaf` should be reverted too

vimwitch: Yes, you're taking the commits you made and applying them to a new base commit (the latest commit of `main`). Copying can work, but usually the only interactive parts of rebasing are handling conflicts. Copying code can work as long as it doesn't break your new functionality.

vimwitch (approved):
lgtm

packages/circuits/package.json
```diff
@@ -33,11 +33,11 @@
     },
     "dependencies": {
         "@ethersproject/solidity": "^5.7.0",
+        "@semaphore-protocol/identity": "^3.5.0",
```
vimwitch, line 36: Actually I should have noticed this before, but this should stay a dev dependency. It's only used in tests.

vimwitch: be762b154d292b5ca7cb8b884c81da998b924f2c

Issue #349, opened 2023-03-03
user skip user state transition but receive attestation


vivianjeng:
It is possible that
1. user A signup
2. epoch transition (`epoch = 1`) (no user state transition)
3. user A generate epoch key with `epoch=1` and receive attest
4. epoch transition (`epoch = 2`) (so the attestation will be included)
5. user A performs user state transition

user should transition from epoch 0 to 1 first
and then transition from 1 to 2
to receive the attestation

and if user A transitions from 0 to 2
then the DB should have an error because the user A doesn't receive the data but it is in `'Attestation'`

so `userStateTransition()` in `Unirep.sol` should not check current epoch
and we should find out user skips which epoch

vimwitch:
So this issue is to prevent `@unirep/core` from picking up an attestation from an epoch the user did not transition to?

I think we shouldn't allow the user to transition to an earlier epoch to claim an attestation. I think they _can_ if they choose, but this breaks non-repudiability. Applications probably shouldn't be attesting to epoch keys without verifying that a user has transitioned to the epoch.

PR #354, opened 2023-03-20
fix(core): user should not receive data if not transition to that epoch

vivianjeng:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #349 

## What is the new behavior?

- Skip epoch that there is no nullifier or signup event
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/UserState.ts
+++ b/packages/core/src/UserState.ts
@@ -211,7 +211,7 @@ export default class UserState {
         const data = Array(this.sync.settings.fieldCount).fill(BigInt(0))
         const orClauses = [] as any[]
         const attesterId = this.sync.attesterId
-        const toEpoch = _toEpoch ?? this.sync.calcCurrentEpoch()
+        const toEpoch = _toEpoch ?? (await this.latestTransitionedEpoch())
         for (let x = 0; x <= toEpoch; x++) {
             const epks = Array(this.sync.settings.numEpochKeyNoncePerEpoch)
                 .fill(null)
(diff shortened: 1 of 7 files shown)


packages/core/src/UserState.ts
```diff
@@ -211,7 +211,7 @@ export default class UserState {
         const data = Array(this.sync.settings.fieldCount).fill(BigInt(0))
         const orClauses = [] as any[]
         const attesterId = this.sync.attesterId
-        const toEpoch = _toEpoch ?? this.sync.calcCurrentEpoch()
+        const toEpoch = _toEpoch ?? (await this.latestTransitionedEpoch())
```
vimwitch, line 214: This breaks the functionality of `getData`. [`getData`](https://developer.unirep.io/docs/next/core-api/user-state#getdata) returns all the data a user has received. [`getProvableData`](https://developer.unirep.io/docs/next/core-api/user-state#getprovabledata) returns the provable data.

vimwitch (changes requested):
Looks good to me, just need to revert the `toEpoch` change

packages/core/src/UserState.ts
vivianjeng: I think it should be `latestTransitionedEpoch`
and it can also be `currentEpoch` if user already transition to the current epoch
so it can make sure the user already transition to this epoch
(if user doesn't transition but receive data, it could be wrong)

e.g. if current epoch is 5 but user's latest transition epoch is 3
and it shouldn't have rep data in epoch 5 if he doesn't transition to 5 
(all data is ~3 epochs and provable data is ~2 epochs)
if he then transitions to 5, it can be valid (all data is ~5 epochs and provable data is ~4 epoch)
if he does not transition to 5, but receive rep data at 5, it should not be valid

vimwitch (approved):

PR #364, opened 2023-03-24
fix(contracts): increase runs value to max

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## What is the new behavior?

The `runs` value in the solidity optimizer is designed to optimize between deploy cost and call cost. Low `runs` value optimizes for a low deployment cost, high runs value optimizes for call costs. `runs` is the number of times the contract can be expected to be called. [More info here](https://docs.soliditylang.org/en/latest/internals/optimizer.html#optimizer-parameter-runs)

In the case of unirep we're deploying a global contract that many applications/users will interact with over time. As a result we should optimize for call cost as much as possible because the deploy cost is a (more or less) one time expenditure.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/contracts/hardhat.config.ts
+++ b/packages/contracts/hardhat.config.ts
@@ -23,7 +23,7 @@ const config: HardhatUserConfig = {
             {
                 version: '0.8.17',
                 settings: {
-                    optimizer: { enabled: true, runs: 200 },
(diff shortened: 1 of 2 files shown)


vivianjeng (commented):
and there is also a `hardhat.config.ts` in core

vivianjeng (approved):

PR #376, opened 2023-03-30
fix(contracts): prevent epoch key 0 and 1 to be attested

### LINKED ISSUE #365: special ordered sorted tree preImages
We should prevent the attestation given to special epoch key `0` and `1`
and prevent the special attestation `[0,0,0,0,0,0,0]` and `[1,0,0,0,0,0,0]`
since it collides with the preserved ordered sorted tree preImage

vivianjeng:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: close #365 

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


packages/contracts/contracts/Unirep.sol
```diff
@@ -257,7 +257,8 @@ contract Unirep is IUnirep, VerifySignature {
             uint currentEpoch = updateEpochIfNeeded(uint160(msg.sender));
             if (epoch != currentEpoch) revert EpochNotMatch();
         }
-        if (epochKey >= SNARK_SCALAR_FIELD) revert InvalidEpochKey();
+        if (epochKey >= SNARK_SCALAR_FIELD || epochKey == 0 || epochKey == 1)
```
vimwitch, line 260: The epoch key 1 is valid, it should instead be `SNARK_SCALAR_FIELD - 1` that is an invalid epoch key.

packages/contracts/contracts/Unirep.sol
```diff
@@ -283,6 +284,7 @@ contract Unirep is IUnirep, VerifySignature {
                 Polysum.add(epkPolysum, vals, EPK_R);
             }
             if (fieldIndex < sumFieldCount) {
+                if (change == 0) revert InvalidAttestation();
```
vimwitch, line 287: We should allow 0 attestations. Practically this is a no-op, but there's nothing explicitly wrong with it, and the economics of the blockchain should naturally de-incentivize applications from doing it.

vimwitch (changes requested):

packages/contracts/contracts/Unirep.sol
vivianjeng: if the epoch key is `1` it will be caught in 
https://github.com/Unirep/Unirep/blob/a3e7510c8c369ebfddbefd66f6a4c5eff757de7d/packages/circuits/circuits/buildOrderedTree.circom#L28-L30

Then it will be recognized as the last leaf and not processing all the other pre-images

Why is `SNARK_SCALAR_FIELD - 1` invalid?

packages/contracts/contracts/Unirep.sol
vivianjeng: I think if there is no limit in attestation it would be fine
but we have `MaxAttestations` limitation so people can send empty attestation and avoid receiving bad attestations or prevent others from receiving attestations.

packages/contracts/contracts/Unirep.sol
vimwitch: You're right, i was thinking about the boundaries of the array. `0` and `1` is correct.

packages/contracts/contracts/Unirep.sol
vivianjeng: it only checks the sum field
so they can attest 0 to a replace field

vimwitch: Yes, that's why I'm saying we don't need this check. It doesn't stop anything.

vivianjeng: but attest 0 to a replace field makes sense
like it can reset the replace field
but attest 0 to a sum field doesn't make change to the balance

vimwitch: Yes it's a no-op, but it's not harmful, I don't think we shouldn't prevent applications from doing this.

vivianjeng: okay I think it is also fine not to handle this for attesters.

vimwitch (approved):

PR #382, opened 2023-04-04
Reveal all epk

### LINKED ISSUE #361: Polysum birthday attack
After some analysis it was found that the polysum/poseidon combination can be birthday attacked in roughly `(8/11)*x` work (where `x` is the number of bits of security for poseidon over `alt_bn128`). This probably puts the security of the polysum scheme at < 90 bits against birthday attacks.

To fix this we should

1. switch the leaf calculation away from polysum and instead use recursive hashing
2. harden the epoch tree polysum logic using something like `block.prevrandao` to reduce the time an attacker has

### LINKED ISSUE #368: Fix circuit tests
https://github.com/Unirep/Unirep/blob/6e635ce203fc7d251432afd52d4281f65f232e1a/packages/circuits/test/userStateTransition.test.ts#L104-L110

It should check the timestamp `changes[i+1]` instead of returning `d` directly

vimwitch:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

Closes #361 
Closes #368 

- [x] Yes
- [ ] No


vimwitch:
Some benchmarks showing UST size for different configurations

```
[INFO]  snarkJS: # of Constraints: 29332

    STATE_TREE_DEPTH: 20,
    EPOCH_TREE_DEPTH: 20,
    EPOCH_TREE_ARITY: 2,
    NUM_EPOCH_KEY_NONCE_PER_EPOCH: 3,
    FIELD_COUNT: 6,
    SUM_FIELD_COUNT: 4,
```


```
[INFO]  snarkJS: # of Constraints: 24472

    STATE_TREE_DEPTH: 15,
    EPOCH_TREE_DEPTH: 15,
    EPOCH_TREE_ARITY: 2,
    NUM_EPOCH_KEY_NONCE_PER_EPOCH: 3,
    FIELD_COUNT: 6,
    SUM_FIELD_COUNT: 4,
```

```
[INFO]  snarkJS: # of Constraints: 34546

    STATE_TREE_DEPTH: 15,
    EPOCH_TREE_DEPTH: 15,
    EPOCH_TREE_ARITY: 2,
    NUM_EPOCH_KEY_NONCE_PER_EPOCH: 3,
    FIELD_COUNT: 14,
    SUM_FIELD_COUNT: 8,
```

packages/core/src/UserState.ts
```diff
@@ -88,14 +87,20 @@ export default class UserState {
         const currentEpoch = await this.sync.loadCurrentEpoch()
         let latestTransitionedEpoch = 0
         for (let x = currentEpoch; x >= 0; x--) {
-            const epkNullifier = genUserStateTransitionNullifier(
-                this.id.secret,
-                this.sync.attesterId.toString(),
-                x
+            const nullifiers = [
```
vimwitch, line 90: I removed the old epoch transition nullifier because we can use the epoch keys that are output by the UST proof. We can use just the first one, it will be either epoch key with nonce 0, or nonce `MAX_NONCE`, depending on if the epoch key received an attestation.

It should be safe to use as the transition nonce because for a certain epoch/epoch tree combination it will always be the same.

vivianjeng (approved):
I like this PR. it is simpler right now.
We can create an issue for adding tests for `ReusableMerkleTree`

PR #386, opened 2023-04-06
Add history tree

### LINKED ISSUE #290: Don't reveal `fromEpoch` during user state transition
Revealing the `fromEpoch` during a UST gives an outside observer a lot of information about the users of the system.

e.g. say that 10 people transition to epoch 4

It's now epoch 10, and only 2 users have transition from epoch 4. We can conclude that 8 users from epoch 4 have not received attestations since epoch 4.

--

We can avoid revealing the `fromEpoch` during UST by building an incremental merkle tree of state tree roots/epoch tree roots of sealed epochs.

Say we have a binary incremental merkle tree with 2^32 leaves. Call this the "history tree". Each leaf of this tree is `H(state tree root, epoch tree root)`. When a user performs a UST instead of revealing an epoch tree root and state tree root they can reveal a history tree root.

This is safe because:
- Each epoch transition nullifier is globally unique
- Each non-empty state tree, and non-empty epoch tree root is globally unique
  - State tree leaves have the current epoch
  - Epoch tree leaves have the epoch key (which changes each epoch)

If we reveal a history tree root we don't need to reveal a `fromEpoch`. Instead it's proven to be a valid pair of state/epoch tree roots which the user has not already transitioned from.

--

To implement this we can add a leaf to the history tree when an epoch is sealed. If an epoch has no attestations and no users in the state tree, no leaf is added to the history tree (e.g. no one should transition from this epoch because no one participated in this epoch).

Using a history tree with 2^32 leaves should allow ~136 years of operation using 1 second epochs.

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Closes #290 
Depends #382 

- [x] Yes
- [ ] No


packages/contracts/contracts/interfaces/IUnirep.sol
```diff
@@ -69,6 +71,7 @@ interface IUnirep {
     error InvalidEpoch(uint256 epoch);
 
     error InvalidProof();
+    error InvalidHistoryTreeRoot(uint256 historyTreeRoot);
     error InvalidStateTreeRoot(uint256 stateTreeRoot);
     error InvalidEpochTreeRoot(uint256 epochTreeRoot);
```
vivianjeng, line 76: The 
```solidity
error InvalidEpochTreeRoot(uint256 epochTreeRoot);
```
can be removed. and remember to update documentation (error codes)
using `yarn contracts errorcode` to get all errors and error codes

vivianjeng (commented):
documentation should be updated: describe history tree and some additional methods e.g. `genHistoryTree`

packages/core/src/Synchronizer.ts
```diff
@@ -515,6 +518,22 @@ export class Synchronizer extends EventEmitter {
         return tree
     }
 
+    async genHistoryTree(): Promise<IncrementalMerkleTree> {
+        const tree = new IncrementalMerkleTree(this.settings.historyTreeDepth)
+        const leaves = await this._db.findMany('HistoryTreeLeaf', {
+            where: {
+                attesterId: this.attesterId.toString(),
+            },
```
vivianjeng, line 526: I prefer that merge #367 before this 
and set `attesterId` as its optional input here

some bugs are detected in #367 that the epoch key index is not universal among each attester
[45a94fa](https://github.com/Unirep/Unirep/pull/367/commits/45a94faa0f29c0fc8ffad1a0dc4afc51bd025b12)

vimwitch: > some bugs are detected in https://github.com/Unirep/Unirep/pull/367 that the epoch key index is not universal among each attester

can you describe this more?

vivianjeng (approved):
LGTM

PR #398, opened 2023-04-07
fix(core): getEpochKeyIndex filter by attesterId

### LINKED ISSUE #397: epoch key index should be queried from certain attester
https://github.com/Unirep/Unirep/blob/f9b93993307435151c3362ea7c76093b0086e67e/packages/core/src/UserState.ts#L385-L387

user state transition will take `epochKeyLeafIndices` from `getEpochKeyIndex`
but it is not unique depends on how many attesters are synced
in this case
The `getEpochKeyIndex` should take an attester ID as input

vimwitch:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Closes #397 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/UserState.ts
+++ b/packages/core/src/UserState.ts
@@ -343,6 +343,7 @@ export default class UserState {
         const attestations = await this.sync._db.findMany('Attestation', {
             where: {
                 epoch,
+                attesterId: this.sync.attesterId.toString(),
             },
             orderBy: {
                 index: 'asc',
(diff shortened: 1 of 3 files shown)


packages/core/src/UserState.ts
```diff
@@ -343,6 +343,7 @@ export default class UserState {
         const attestations = await this.sync._db.findMany('Attestation', {
             where: {
                 epoch,
+                attesterId: this.sync.attesterId.toString(),
```
vivianjeng, line 346: I prefer that this can be a input in 
```ts
public getEpochKeyIndex = async (
     epoch: number,
     _epochKey: bigint | string
 )
```
so that it can support multi-attesters synchronizer/user state

vimwitch: I think this should be done either in the multi-attester pr, or by basing this branch on the multi-attester pr. Those changes aren't in main yet

PR #402, opened 2023-04-16
337 replacement counter field

### LINKED ISSUE #337: Replacement field timestamp
A timestamp is used to determine the order of replacement fields. It's possible that the timestamp is the same for two transactions, and thus 2 attestations. This would result in indeterminate ordering (it would be based on which epoch key nonce is lower/higher).

Instead of using a timestamp we can using a `uint` counter in the contract and increment it with every attestation.

chndr1:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Replacement fields for user state transition are calculated using timestamps. 

Closes #337 

## What is the new behavior?

The upper 48 bits of the `change`  field in the userStateTransition circuit contain the current attestation count. This counter is incremented on every attestation. This ensures that every attestation has a unique number that can be used to order it. 

- [ ] Yes
- [x] No

## Other information
N/A


packages/contracts/contracts/Unirep.sol
```diff
@@ -69,6 +71,8 @@ contract Unirep is IUnirep, VerifySignature {
         epochKeyVerifier = _epochKeyVerifier;
         epochKeyLiteVerifier = _epochKeyLiteVerifier;
 
+        attestationCounter = 0;
```
vimwitch, line 74: No need to initialize to 0, all variables default to 0 in the evm.

packages/contracts/contracts/Unirep.sol
```diff
@@ -112,6 +116,7 @@ contract Unirep is IUnirep, VerifySignature {
                 x,
                 initialData[x]
             );
+            attestationCounter++;
```
vimwitch, line 119: No need to increment the counter here. During signup any custom data must use nonce 0. This is enforced on line 110, we check that all the upper bits are 0.

packages/contracts/contracts/Unirep.sol
```diff
@@ -288,6 +293,7 @@ contract Unirep is IUnirep, VerifySignature {
             fieldIndex,
             change
         );
+        attestationCounter++;
```
vimwitch, line 296: I think it's ok to increment this value for all attestations. Technically we only need to increment this value when an attestation is made to a replacement field. But we'll only hit 2^48 attestations after 9 years of 1 million attestations per second.

Plus i think it's interesting to have an onchain variable tracking the attestation count for the system.

packages/contracts/test/attestation.test.ts
```diff
@@ -146,12 +146,10 @@ describe('Attestations', function () {
         const epochKey = BigInt(24910)
```
vimwitch, line 146: We need one more test case

- Attest to a replacement field and check that the data matches the expected value

Check this in the emitted event for the attestation transaction

vimwitch (changes requested):

packages/contracts/contracts/Unirep.sol
chndr1: In this case, I'll leave it as it is

packages/contracts/test/attestation.test.ts
chndr1: Added new test case to attest with different values and verify that it is the expected value

packages/contracts/contracts/Unirep.sol
```diff
@@ -46,6 +46,8 @@ contract Unirep is IUnirep, VerifySignature {
     uint8 public immutable numEpochKeyNoncePerEpoch;
     uint8 public immutable replNonceBits;
 
+    uint48 public attestationCounter;
```
vimwitch, line 49: I think let's call this `attestationCount`. I think it makes more sense when accessing it.

packages/contracts/test/attestation.test.ts
```diff
@@ -204,4 +203,80 @@ describe('Attestations', function () {
...
+
+        const epochKey = BigInt(24910)
+        let attestationCounter = await unirepContract.attestationCounter()
+
+        const fieldIndex = SUM_FIELD_COUNT
+
+        for (let x = 1; x <= 3; x++) {
+            const tx = await unirepContract
+                .connect(attester)
+                .attest(epochKey, epoch, fieldIndex, x)
```
vimwitch, line 261: This is a good test, and I think the logic is right, but it would be nice to attest to a random value. This would guarantee that `x != nonce`, which would cause a false positive.

You can get a random value from `randomf`

```js
import randomf from 'randomf'

const v = randomf(BigInt(2)**BigInt(254-REPL_NONCE_BITS))
```

vimwitch (changes requested):
looks good, couple small things

packages/contracts/test/attestation.test.ts
chndr1: Fixed, using random value for nonce in the test

vimwitch (approved):

PR #416, opened 2023-04-20
feat(contracts): lazy merkle tree, reusable merkle tree tests

### LINKED ISSUE #389: Add tests for `ReusableMerkleTree`

### LINKED ISSUE #411: Lazily build merkle tree roots onchain
Right now when we add an item to a merkle tree we calculate the root immediately. This means every addition takes `HEIGHT` hashes, regardless of how full the tree is.

Instead we can build the smallest possible subtree upon insertion. For example, when we insert the first element we simply do `H(element, 0)` to calculate the first parent. If we want to access the root we'll then hash up the height of the tree to get the root.

As we insert elements the complexity increases until the tree is almost full, at which point we have the same complexity as a normal implementation. The amortized cost should be ~25% less than a normal implementation. Early insertions/updates in the tree are _much_ cheaper, with later insertions costing the same as a normal implementation.

This works well for trees that are changed often, but rarely have their root accessed. This is good for the epoch tree in attesters. This tree is changed often, but the root only needs to be accessed after an epoch ends, at which point the tree is immutable (so the final root can be calculated once and stored).

A visual of adding elements to a tree of depth 3

Insertion cost: 0 hashes
```
a
```

Insertion cost: 1 hash
```
  r1
 /  \
a    b
```

Insertion cost: 2 hashes
```
      r3
   /       \
  s0       s1
 /  \    /    \
a   b   c      0
```

Insertion cost: 2 hashes
```
      r4
   /       \
  s0       s1
 /  \    /    \
a   b   c      d
```

Insertion cost: 3 hashes (s3 is a constant here)
```
              r5
        /              \
      s4                s5
   /       \         /        \
  s0       s1       s2        s3
 /  \    /    \    /   \    /    \
a   b  c       d   e   0    0      0
```

Insertion cost: 3 hashes (s3 is a constant here)
```
              r6
        /              \
      s4                s5
   /       \         /        \
  s0       s1       s2        s3
 /  \    /    \    /   \    /    \
a   b  c       d   e   f    0      0
```

Insertion cost: 3 hashes (s3 is no longer constant, but s2 is)
```
              r6
        /              \
      s4                s5
   /       \         /        \
  s0       s1       s2        s3
 /  \    /    \    /   \    /    \
a   b  c       d   e   f    g      0
```

For a tree of depth 3 a normal implementation needs 24 hashes to fill the tree. Using the lazy approach we only need 17 hashes.

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Closes #411 
Closes #389 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/contracts/contracts/libraries/LazyMerkleTree.sol
+++ b/packages/contracts/contracts/libraries/LazyMerkleTree.sol
@@ -2,11 +2,13 @@
 pragma solidity ^0.8.0;
 
 import 'poseidon-solidity/PoseidonT3.sol';
+import 'hardhat/console.sol';
 
 struct LazyTreeData {
     uint32 maxIndex;
     uint40 numberOfLeaves;
     mapping(uint256 => uint256) elements;
+    uint32[32] levelCounts;
 }
 
 library LazyMerkleTree {
@@ -23,6 +25,9 @@ library LazyMerkleTree {
 
     function reset(LazyTreeData storage self) public {
         self.numberOfLeaves = 0;
+        for (uint8 x = 0; x < MAX_DEPTH; x++) {
+            self.levelCounts[x] = 0;
+        }
     }
 
     function indexForElement(
@@ -46,6 +51,9 @@ library LazyMerkleTree {
 
         for (uint8 i = 0; true; ) {
             self.elements[indexForElement(i, index)] = hash;
+            if (i > 0) {
+                self.levelCounts[i] = uint32(index) + 1;
+            }
             // it's a left element so we don't hash until there's a right element
             if (index & 1 == 0) break;
             uint40 elementIndex = indexForElement(i, index - 1);
@@ -77,10 +85,11 @@ library LazyMerkleTree {
                 levels[i + 1] = PoseidonT3.hash([levels[i], defaultZero(i)]);
             } else {
                 // if the element is a right sibling the parent is guaranteed to be calculated
-                uint256 parent = self.elements[
-                    indexForElement(i + 1, index >> 1)
-                ];
-                if (parent != 0) {
+                uint256 levelCount = self.levelCounts[i + 1];
+                if (levelCount > index >> 1) {
+                    uint256 parent = self.elements[
+                        indexForElement(i + 1, index >> 1)
+                    ];
                     levels[i + 1] = parent;
                 } else {
                     uint256 sibling = self.elements[
--- a/packages/contracts/contracts/libraries/LazyMerkleTree.sol
+++ b/packages/contracts/contracts/libraries/LazyMerkleTree.sol
@@ -57,6 +57,38 @@ library LazyMerkleTree {
         }
     }
 
+    function update(
+        LazyTreeData storage self,
+        uint256 leaf,
+        uint40 index
+    ) public {
+        require(
+            leaf < SNARK_SCALAR_FIELD,
+            'LazyMerkleTree: leaf must be < SNARK_SCALAR_FIELD'
+        );
+        uint40 numberOfLeaves = self.numberOfLeaves;
+        require(index < numberOfLeaves);
+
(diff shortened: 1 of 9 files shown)


vimwitch:
The insertion cost for the lazy merkle tree is nearly constant with depth. The cost of calculating the root is (in the worst case) the same as inserting into a normal tree. The lazy merkle tree is also flexible depth, a root can be calculated for any depth, so as the tree expands the depth can logically be increased without changing the data structure.

For trees that are append only proofs can be made against subtrees to avoid the cost/race condition of calculating the latest root.

### Benchmarks

Filling a tree of depth 5
```
Average insertion cost (lazy): 77,096
Average insertion cost (normal): 230,921
```

Filling a tree of depth 9
```
Average insertion cost (lazy): 84,668
Average insertion cost (normal): 387,346
```

Filling a tree of depth 10
```
Average cost (lazy): 85,080
Average cost (normal): 428,364
```

Inserting 128 leaves into a tree of depth 32
```
Average insertion cost (lazy): 85,185
Average insertion cost (normal): 1,386,428
```

vimwitch:
Actually the insertion cost is constant, the amortized cost for a tree just changes as the number of leaves inserted changes.

packages/contracts/contracts/libraries/LazyMerkleTree.sol
```diff
@@ -0,0 +1,202 @@
...
+            // it's a left element so we don't hash until there's a right element
+            if (index & 1 == 0) break;
+            uint40 elementIndex = indexForElement(i, index - 1);
+            hash = PoseidonT3.hash([self.elements[elementIndex], hash]);
+            unchecked {
+                index >>= 1;
+                i++;
+            }
+        }
+    }
```
vimwitch, line 58: Need to add an update function

packages/contracts/contracts/libraries/LazyMerkleTree.sol
```diff
@@ -0,0 +1,202 @@
...
+
+        for (uint8 i = 0; i < depth; ) {
+            if (index & 1 == 0) {
+                levels[i + 1] = PoseidonT3.hash([levels[i], defaultZero(i)]);
+            } else {
+                // if the element is a right sibling the parent is guaranteed to be calculated
+                uint256 parent = self.elements[
+                    indexForElement(i + 1, index >> 1)
+                ];
+                if (parent != 0) {
```
vimwitch, line 83: Don't rely on a zero check for the parent if the tree is reusable. Instead track how many elements have been inserted on each level.

vimwitch:
Updating a lazy merkle tree is 33% cheaper in the _worst_ case. If the tree is less full it's even cheaper.

Updating a tree of depth 6 repeatedly as it's filled
```
Average update cost (lazy): 155,399
Average update cost (normal): 240,408
```

These are benchmarks for attestations using the same configuration (epoch tree depth = 17).

Normal tree
```bash
·············|··················|·············|·············|····················|·············|················
|  Contract  ·  Method          ·  Min        ·  Max        ·  Avg               ·  # calls    ·  eur (avg)    │
·············|··················|·············|·············|····················|·············|················
|  Unirep    ·  attest          ·     349662  ·    1237754  ·            511204  ·         37  ·            -  │
·············|··················|·············|·············|····················|·············|···············
```

Lazy tree
```bash
·············|··················|·············|·············|····················|·············|················
|  Contract  ·  Method          ·  Min        ·  Max        ·  Avg               ·  # calls    ·  eur (avg)    │
·············|··················|·············|·············|····················|·············|················
|  Unirep    ·  attest          ·     255880  ·     418455  ·            320846  ·         33  ·            -  │
·············|··················|·············|·············|····················|·············|···············
```

vivianjeng (approved):
LGTM
Thank you 🙏🏻

PR #427, opened 2023-04-24
357 separate proof verification

### LINKED ISSUE #357: Separate proof verification functions from UniRep contract
The UniRep contract has exceeded 24 kb, which is the limit for deploying to most chains. We should separate some of the proofs and their associated verification functions into separate contracts.

The `reputationProof`, `epochKeyProof` and `epochKeyLite` verification functions (along with signal encoding/decoding) can be moved out of the `Unirep` contract and into their own contract.

chndr1:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Currently, Epoch Key Proof Verification, Reputation Proof Verification, and Epoch Key Proof Lite Verification functions all occur in the Unirep contract. This means that the Unirep contract is approximately 18,000 bytes large.

Closes #357 

## What is the new behavior?
The Epoch Key Proof Verification, Reputation Proof Verification, and Epoch Key Proof Lite Verification functions have all been split into different contracts and deployed individually. This means that the Unirep contract size is now only 14,406 bytes large. Additionally, this means that the caller now has the responsibility of supplying the current epoch and state tree to verifiers.

- [x] Yes
- [ ] No

This change affects how the Unirep contract is deployed. It is now deployed independently of its verifiers.


packages/contracts/contracts/EpochKeyLiteProofVerifier.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {IUnirep} from './interfaces/IUnirep.sol';
+import {IVerifier} from './interfaces/IVerifier.sol';
+
+import {EpochKeyProofVerifier} from './EpochKeyProofVerifier.sol';
+
+contract EpochKeyLiteProofVerifier is IUnirep {
```
vimwitch, line 9: This contract shouldn't implement `IUnirep`. `IUnirep` defines storage variables for the unirep contract - the verifier doesn't need this storage.

packages/contracts/contracts/EpochKeyLiteProofVerifier.sol
```diff
@@ -0,0 +1,56 @@
...
+    IVerifier public epochKeyLiteVerifier;
+    EpochKeyProofVerifier public epochKeyProofVerifier;
```
vimwitch, line 11: In Solidity there's a [large-ish cost](https://eips.ethereum.org/EIPS/eip-2929#specification) to accessing other contracts in a call. Instead of storing a reference to the `EpochKeyProofVerifier` it would be better to define a base class that implements `decodeEpochKeyControl`, and then have both `EpochKeyProofVerifier` and `EpochKeyLiteProofVerifier` inherit the base contract. Then both instances have their own copy of `decodeEpochKeyControl`.

packages/contracts/contracts/EpochKeyProofVerifier.sol
```diff
@@ -0,0 +1,79 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {IUnirep} from './interfaces/IUnirep.sol';
+import {IVerifier} from './interfaces/IVerifier.sol';
+
+contract EpochKeyProofVerifier is IUnirep {
```
vimwitch, line 7: Also don't implement `IUnirep` here.

packages/contracts/contracts/EpochKeyLiteProofVerifier.sol
vimwitch: Ah I see you imported to get access to the errors we define in `IUnirep`. Instead you should define something like `IProofVerifier` or `IBaseVerifier` that defines the errors you need as well as maybe `SNARK_SCALAR_FIELD`

packages/contracts/contracts/EpochKeyProofVerifier.sol
```diff
@@ -0,0 +1,79 @@
...
+            signals.nonce
+        ) = decodeEpochKeyControl(publicSignals[2]);
+        return signals;
+    }
+
+    function verifyEpochKeyProof(
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof,
+        uint48 attesterCurrentEpoch,
+        bool attesterStateTreeRootExists
```
vimwitch, line 57: Instead of taking this boolean as an input i think the caller should manually make this check. This function should return `(uint48 epoch, uint256 stateTreeRoot)`, or better yet, just return `EpochKeySignals`.

The caller will likely want to apply additional checks. In this function we want to apply the bare minimum checks that _always_ need to be done.

packages/contracts/contracts/ReputationProofVerifier.sol
```diff
@@ -0,0 +1,100 @@
...
+            proveZeroRep,
+            proveGraffiti
+        );
+    }
+
+    function verifyReputationProof(
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof,
+        uint48 attesterCurrentEpoch,
+        bool attesterStateTreeRootExists
```
vimwitch, line 84: Same here, should return a `ReputationSignals` object for further verification.

packages/contracts/contracts/EpochKeyLiteProofVerifier.sol
vimwitch: Also the structs `EpochKeySignals` and `ReputationSignals` can be defined in this interface.

vimwitch: Instead of making it an interface it could be a base contract that these proof verifiers inherit. Then you could define the `constructor` implementation as well (accept a single variable and assign it to a state variable `IVerifier verifier`).

packages/contracts/test/verifier.test.ts
```diff
@@ -56,10 +61,16 @@ describe('Epoch key lite proof verifier', function () {
     this.timeout(300000)
 
     let unirepContract
+    let epochKeyLiteProofVerifier
+    let epochKeyProofVerifier
 
     before(async () => {
         const accounts = await ethers.getSigners()
         unirepContract = await deployUnirep(accounts[0])
+        epochKeyLiteProofVerifier = await deployEpochKeyLiteProofVerifier(
```
vimwitch, line 70: The verifier contracts are decoupled from the unirep contract in this PR. There's no need to deploy `unirepContract` in these tests. You can make proofs with arbitrary inputs to test the verifiers.

This means you won't use the `UserState` object, you'll directly use `defaultProver` with whatever inputs you want.

packages/core/test/synchronizerAttesters.test.ts
```diff
@@ -329,9 +334,20 @@ describe('Synchronizer watch multiple attesters', function () {
...
-                .verifyEpochKeyProof(publicSignals, proof)
-                .then((t) => t.wait())
+            const root = unirepContract.attesterStateTreeRoot(attesterId)
+            const attesterStateTreeRootExists =
+                unirepContract.attesterStateTreeRootExists(
+                    attesterId,
+                    epoch,
+                    root
+                )
+            await epochKeyProofVerifier.verifyEpochKeyProof(
```
vimwitch, line 344: I think this test is unnecessary?

packages/contracts/contracts/EpochKeyProofVerifier.sol
```diff
@@ -0,0 +1,79 @@
...
+        (
+            signals.revealNonce,
+            signals.attesterId,
+            signals.epoch,
+            signals.nonce
+        ) = decodeEpochKeyControl(publicSignals[2]);
+        return signals;
+    }
+
+    function verifyEpochKeyProof(
```
vimwitch, line 53: This function could probably be named more generically. In the calling contract it will likely be stored as a variable like:

```solidity
EpochKeyProofVerifier epochKeyVerifier;
```

If the function is named `verifyEpochKeyProof` you get

```solidity
epochKeyVerifier.verifyEpochKeyProof(...);
```

Instead it's okay to name it like `verifyAndCheck`. Then all the verifiers can have similar function names. In the calling code it should look like

```solidity
epochKeyVerifier.verifyAndCheck(...);
```

vimwitch, line 53: It might be good to include an additional function that checks the `attesterId` variable. This could be like `verifyAndCheckCaller`. This would perform all the checks here, but also have `require(attesterId == msg.sender)`.

This should be a separate function because the `msg.sender` will not _always_ be the attester.

packages/contracts/contracts/Unirep.sol
```diff
@@ -9,6 +9,10 @@ import {VerifySignature} from './libraries/VerifySignature.sol';
 import {IUnirep} from './interfaces/IUnirep.sol';
 import {IVerifier} from './interfaces/IVerifier.sol';
 
+import {ReputationProofVerifier} from './ReputationProofVerifier.sol';
```
vimwitch, line 12: No need to import these.

vimwitch (changes requested):
Contracts are looking nicely decoupled, things can be simplified in a few places.

packages/contracts/test/verifier.test.ts
chndr1: In this case, how would you prove that an attester state tree root exists without having access to the `unirepContract` AttesterData

packages/contracts/contracts/EpochKeyProofVerifier.sol
chndr1: Refactored

packages/contracts/contracts/Unirep.sol
chndr1: Deleted

packages/contracts/contracts/ReputationProofVerifier.sol
chndr1: Change added

packages/contracts/contracts/EpochKeyProofVerifier.sol
chndr1: Should this be the function that is called by default? Also, should I add tests for this?

packages/contracts/test/verifier.test.ts
vimwitch: We're decoupling the verifier and unirep contract. In the epoch key contract for example we only need to verify that it correctly decodes the control output and returns the variables from the public signals. We don't actually care if the state tree root is valid, just that it's correctly parsed from the proof. State root validity is checked in the other tests.

The verifier isn't responsible for checking the state root, just for returning it.

packages/contracts/contracts/EpochKeyProofVerifier.sol
vimwitch: I don't think either of them need to be called by default. Either can be used by someone that wants to verify.

To test you should generate a proof with an attester id that is a contract address you control, e.g.

```js
const [owner] = await ethers.getAccounts()
```

You'd make a proof with `owner.address` as the attester id and make sure it doesn't revert (when you called `verifyAndCheckCaller` from `owner`. Then make another proof with a random attester id and make sure that it reverts when called with `owner`.

packages/contracts/contracts/BaseVerifier.sol
```diff
@@ -0,0 +1,80 @@
...
+
+contract BaseVerifier {
+    IVerifier verifier;
+
+    uint256 public constant SNARK_SCALAR_FIELD =
+        21888242871839275222246405745257275088548364400416034343698204186575808495617;
+
+    struct EpochKeySignals {
+        uint256 revealNonce;
+        uint256 stateTreeRoot;
```
vimwitch, line 14: These don't need to all be `uint256`. Try to use the smallest types possible so the EVM will use less memory to represent the struct.

e.d. `revealNonce` can be a `bool`, `nonce` can be a `uint8`, `attesterId` can be `uint160`, `epoch` `uint48`.

packages/contracts/contracts/BaseVerifier.sol
```diff
@@ -0,0 +1,80 @@
...
+        uint256 attesterId;
+        uint256 revealNonce;
+        uint256 proveMinRep;
+        uint256 proveMaxRep;
+        uint256 proveZeroRep;
+        uint256 minRep;
+        uint256 maxRep;
+    }
+
+    error UserAlreadySignedUp(uint256 identityCommitment);
```
vimwitch, line 38: Some of these errors are not used in these contracts. Try to only include errors that are needed.

packages/contracts/contracts/EpochKeyLiteProofVerifier.sol
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {IVerifier} from './interfaces/IVerifier.sol';
+import {BaseVerifier} from './BaseVerifier.sol';
+
+contract EpochKeyLiteProofVerifier is BaseVerifier {
+    constructor(IVerifier _verifier) BaseVerifier(_verifier) {}
```
vimwitch, line 8: You should double check, but I think you shouldn't need to explicitly define this constructor. Try just removing this line and seeing if it works as expected.

packages/contracts/test/verifier.test.ts
```diff
@@ -199,17 +169,19 @@ describe('Epoch key lite proof verifier', function () {
...
             r.proof
         )
-        await expect(
-            unirepContract.verifyEpochKeyLiteProof(publicSignals, proof)
-        ).to.be.revertedWithCustomError(unirepContract, 'InvalidEpoch')
+        const signals = await epochKeyLiteProofVerifier.verifyAndCheck(
+            publicSignals,
+            proof
+        )
+        expect(signals.epoch).to.be.greaterThan(epoch)
```
vimwitch, line 176: You should check that the signals match the expected value. e.g. `signals.epoch == invalidEpoch`. Also check that other outputs like `revealNonce`, `attesterId`, etc.

packages/contracts/test/verifier.test.ts
```diff
@@ -479,61 +565,84 @@ describe('Epoch key proof verifier', function () {
...
-        ).to.be.revertedWithCustomError(unirepContract, 'InvalidStateTreeRoot')
+            const { publicSignals, proof } = new EpochKeyProof(
+                r.publicSignals,
+                r.proof
+            )
+            await expect(
+                epochKeyProofVerifier
+                    .connect(owner)
+                    .verifyAndCheckCaller(publicSignals, proof)
+            ).to.be.reverted
```
vimwitch, line 644: When you check that a call is reverted you should verify that it's reverted with the correct error. See [here](https://github.com/Unirep/Unirep/blob/main/packages/contracts/test/attestation.test.ts#L48) for an example.

packages/core/test/synchronizerAttesters.test.ts
```diff
@@ -329,9 +334,7 @@ describe('Synchronizer watch multiple attesters', function () {
             const attesterId = BigInt(accounts[i].address).toString()
             const { publicSignals, proof, epochKey, epoch } =
                 await userState.genEpochKeyProof({ attesterId })
-            await unirepContract
-                .verifyEpochKeyProof(publicSignals, proof)
-                .then((t) => t.wait())
+            unirepContract._updateEpochIfNeeded(attesterId)
```
vimwitch, line 337: I don't think this line is needed. The `attest` function will call this. It's also not `await`ed

packages/core/test/synchronizerAttesters.test.ts
```diff
@@ -14,10 +17,12 @@ describe('Synchronizer watch multiple attesters', function () {
     this.timeout(0)
 
     let unirepContract: Unirep
+    let epochKeyProofVerifier
 
     before(async () => {
         const accounts = await ethers.getSigners()
         unirepContract = await deployUnirep(accounts[0])
+        epochKeyProofVerifier = await deployEpochKeyProofVerifier(accounts[0])
```
vimwitch, line 25: I think this can be removed?

vimwitch (changes requested):
Looking good, make sure to resolve the conflicts with `main`. I'd recommend doing this by [rebasing](https://git-scm.com/book/en/v2/Git-Branching-Rebasing). During a rebase you'll take the changes you've made on this branch (your commits) and replay them on top of the latest `main` branch. With each commit you'll potentially need to resolve conflicts.

To do this make sure you've pulled the latest `main`

```sh
git checkout main
git pull
git checkout 357-separate-proof-verification
git rebase main
```

It will then start replaying commits and halt if conflicts are found. If you mess something up you can restart by doing `git rebase --abort` and then starting again.

packages/contracts/contracts/BaseVerifier.sol
chndr1: Done, `EpochKeySignals` and `ReputationSignals` now uses the smallest types

packages/contracts/contracts/BaseVerifier.sol
chndr1: Done, remove all unused errors from `BaseVerifier`

packages/contracts/contracts/EpochKeyLiteProofVerifier.sol
chndr1: The contract doesn't compile if the constructor isn't explicitly defined

packages/contracts/test/verifier.test.ts
chndr1: Done, checking for revert message

packages/core/test/synchronizerAttesters.test.ts
chndr1: Deleted `unirepContract._updateEpochIfNeeded(attesterId)`

packages/core/test/synchronizerAttesters.test.ts
chndr1: Yep, this isn't required.

packages/contracts/contracts/proofVerifiers/BaseProofVerifier.sol
```diff
@@ -0,0 +1,64 @@
...
+import {IVerifier} from '../interfaces/IVerifier.sol';
+
+contract BaseProofVerifier {
+    IVerifier verifier;
+
+    uint256 public constant SNARK_SCALAR_FIELD =
+        21888242871839275222246405745257275088548364400416034343698204186575808495617;
+
+    struct EpochKeySignals {
+        bool revealNonce;
```
vimwitch, line 13: Try to put this `bool` near the other small fields. In the EVM each word is 256 bits. When possible the compiler will combine nearby fields into a single work. A `uint8`, `uint48`, `uint160` and `bool` can all be in 1 word.

packages/contracts/contracts/proofVerifiers/EpochKeyLiteProofVerifier.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    function decodeEpochKeyLiteSignals(
+        uint256[] calldata publicSignals
+    ) public pure returns (EpochKeySignals memory) {
+        EpochKeySignals memory signals;
+        signals.epochKey = publicSignals[1];
+        signals.data = publicSignals[2];
+        // now decode the control values
+        (
+            signals.revealNonce,
```
vimwitch, line 18: Didn't know you could destructure directly into a memory variable, that's cool.

packages/contracts/contracts/proofVerifiers/EpochKeyLiteProofVerifier.sol
```diff
@@ -0,0 +1,57 @@
...
+        if (signals.attesterId >= type(uint160).max) revert AttesterInvalid();
+
+        return signals;
+    }
+
+    function verifyAndCheckCaller(
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof
+    ) public view returns (EpochKeySignals memory) {
+        EpochKeySignals memory signals = decodeEpochKeyLiteSignals(
```
vimwitch, line 46: You're double decoding the signals here. Once on line 46, and once on line 55. I think the range checks for `epochKey` and `attesterId` could be moved to the `decodeSignals` function. Then you don't need to worry about calling into `verifyAndCheck` here.

packages/contracts/contracts/proofVerifiers/EpochKeyProofVerifier.sol
```diff
@@ -0,0 +1,55 @@
...
+        uint256[8] calldata proof
+    ) public view returns (EpochKeySignals memory) {
+        EpochKeySignals memory signals = decodeEpochKeySignals(publicSignals);
+
+        require(
+            signals.attesterId == uint160(msg.sender),
+            'attesterId is not caller'
+        );
+
+        return verifyAndCheck(publicSignals, proof);
```
vimwitch, line 53: Double decode

packages/contracts/contracts/proofVerifiers/ReputationProofVerifier.sol
```diff
@@ -0,0 +1,98 @@
...
+
+    function decodeReputationControl(
+        uint256 control
+    )
+        public
+        pure
+        returns (
+            uint256 minRep,
+            uint256 maxRep,
+            uint256 proveMinRep,
```
vimwitch, line 44: These fields can be smaller. `prove*` can each be `bool` values.

packages/contracts/contracts/proofVerifiers/BaseProofVerifier.sol
```diff
@@ -0,0 +1,64 @@
...
+        public
+        pure
+        returns (
+            bool revealNonce,
+            uint160 attesterId,
+            uint48 epoch,
+            uint8 nonce
+        )
+    {
+        revealNonce = (control >> 232) & 1 != 0;
```
vimwitch, line 54: The order of operations here is ambiguous (e.g. `1 != 0` might always evaluate to `1` as the rhs to the `&`). Add parentheses or use an explicit cast instead.

packages/contracts/contracts/proofVerifiers/ReputationProofVerifier.sol
```diff
@@ -0,0 +1,98 @@
...
+        ReputationSignals memory signals = decodeReputationSignals(
+            publicSignals
+        );
+
+        require(
+            signals.attesterId == uint160(msg.sender),
+            'attesterId is not caller'
+        );
+
+        return verifyAndCheck(publicSignals, proof);
```
vimwitch, line 96: Double decode

vimwitch (changes requested):

packages/docs/docs/contracts-api/base-proof-verifier.md
```diff
@@ -0,0 +1,21 @@
...
+## decodeEpochKeyControl
+
+Decode a epoch key related control from [epoch key lite proof](../circuits-api/circuits#epoch-key-lite-proof), [epoch key proof](../circuits-api/circuits.md#epoch-key-proof), and [reputation proof](../circuits-api/circuits.md#prove-reputation-proof) into named variables.
+
+```sol
+function decodeEpochKeyControl(uint256 control)
+    public
+    pure
+    returns (
+      uint256 revealNonce,
```
vimwitch, line 16: These return types are incorrect.

packages/docs/docs/contracts-api/epoch-key-lite-proof-verifier.md
```diff
@@ -0,0 +1,64 @@
...
+    uint256 epochKey;
+    uint256 data;
+}
+```
+
+## verifyAndCheck
+
+Verify an [epoch key lite proof](../circuits-api/circuits#epoch-key-lite-proof) and validate the public signals against the onchain state. This function will revert if any inputs are invalid.
+
+:::caution
```
vimwitch, line 38: Here we should also add a `:::danger:::` note about verifying the `attesterId`. We should recommend using `verifyAndCheckCaller` or manually checking the `attesterId`. Same with other proof contracts.

packages/docs/docs/contracts-api/base-proof-verifier.md
chndr1: Return types changed

packages/docs/docs/contracts-api/epoch-key-lite-proof-verifier.md
chndr1: `verifyAndCheckCaller` recommendation added

vimwitch (approved):

packages/contracts/contracts/proofVerifiers/EpochKeyProofVerifier.sol
```diff
@@ -0,0 +1,61 @@
...
+    function verifyAndCheckCaller(
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof
+    ) public view returns (EpochKeySignals memory) {
+        EpochKeySignals memory signals = decodeEpochKeySignals(publicSignals);
+
+        require(
+            signals.attesterId == uint160(msg.sender),
+            'attesterId is not caller'
+        );
```
vivianjeng, line 53: I prefer not to mix use of `require` and `revert`

packages/contracts/contracts/proofVerifiers/EpochKeyLiteProofVerifier.sol
```diff
@@ -0,0 +1,63 @@
...
+            signals.attesterId == uint160(msg.sender),
+            'attesterId is not caller'
+        );
+
+        bool valid = verifier.verifyProof(publicSignals, proof);
+
+        if (!valid) revert InvalidProof();
+
+        return signals;
+    }
```
vivianjeng, line 62: I think it can reuse `verifyAndCheck`
like
```solidity
    function verifyAndCheckCaller(
        uint256[] calldata publicSignals,
        uint256[8] calldata proof
    ) public view returns (EpochKeySignals memory) {
        require(
            signals.attesterId == uint160(msg.sender),
            'attesterId is not caller'
        );
        return verifyAndCheck(publicSignals, proof)
    }
```

chndr1: Where would `signals` be defined?

vivianjeng: ah right
or like this
```solidity
    function verifyAndCheckCaller(
        uint256[] calldata publicSignals,
        uint256[8] calldata proof
    ) public view returns (EpochKeySignals memory) {
        EpochKeySignals memory signals = verifyAndCheck(publicSignals, proof);
        require(
            signals.attesterId == uint160(msg.sender),
            'attesterId is not caller'
        );
        return signals;
    }
```

chndr1: Change to `verifyAndCheckCaller` added

packages/contracts/contracts/proofVerifiers/BaseProofVerifier.sol
```diff
@@ -0,0 +1,65 @@
...
+        bool proveMinRep;
+        bool proveMaxRep;
+        bool proveZeroRep;
+        bool revealNonce;
+        uint8 nonce;
+        uint48 epoch;
+        uint160 attesterId;
+        uint256 stateTreeRoot;
+        uint256 epochKey;
+        uint256 graffitiPreImage;
```
vivianjeng, line 33: there is no `graffitiPreImage` anymore (just graffiti itself)
https://github.com/Unirep/Unirep/pull/422

vimwitch: @chndr1 This signal still exists it just needs to be renamed to `graffiti`

chndr1: whoops lol. renamed

packages/docs/docs/contracts-api/reputation-proof-verifier.md
```diff
@@ -0,0 +1,91 @@
...
+    bool proveMinRep;
+    bool proveMaxRep;
+    bool proveZeroRep;
+    bool revealNonce;
+    uint8 nonce;
+    uint48 epoch;
+    uint160 attesterId;
+    uint256 stateTreeRoot;
+    uint256 epochKey;
+    uint256 graffitiPreImage;
```
vivianjeng, line 48: The `graffitiPreImage` should be renamed

packages/docs/docs/contracts-api/unirep-sol.md
```diff
@@ -164,7 +164,7 @@ function decodeReputationSignals(uint256[] memory publicSignals)
 struct ReputationSignals {
     uint256 stateTreeRoot;
     uint256 epochKey;
-    uint256 graffiti;
+    uint256 graffitiPreImage;
```
vivianjeng, line 167: The `graffitiPreImage` should be renamed

vivianjeng (commented):
The error codes should be updated as well
(either in this PR or in another PR)
you can use `yarn contracts errorcode` to list all error codes and update in [this file](https://github.com/Unirep/Unirep/blob/357-separate-proof-verification/packages/docs/docs/contracts-api/errors.md)

chndr1:
@vivianjeng One thing that we may want to consider is adding some functionality to read error codes from `BaseProofVerifier.sol` since `errorCodes.js` only reads from the Unirep ABI. This will prevent having to add unrelated error codes to the `IUnirep.sol` interface. I added one (`CallerInvalid()`) but this should be reconsidered in the future.

vivianjeng:
I notice that the `Unirep.sol` does not use the `decodeEpochKeyLiteSignals` from verifiers.
and we define the `EpochKeySignals` and  `ReputationSignals` twice?

Is it possible that `Unirep.sol` can use the struct defined in `BaseProofVerifier`? 
e.g. `SNARK_SCALAR_FIELD`, `EpochKeySignals`, `ReputationSignals` and error codes?

packages/contracts/deploy/deploy.ts
```diff
@@ -27,6 +33,81 @@ const retryAsNeeded = async (fn: any, maxRetry = 10) => {
     }
 }
 
+export const deployVerifiers = async (
+    deployer: ethers.Signer,
+    settings: CircuitConfig = CircuitConfig.default,
```
vivianjeng, line 38: actually the `settings` is not used in this function.

packages/contracts/deploy/deploy.ts
```diff
@@ -27,6 +33,81 @@ const retryAsNeeded = async (fn: any, maxRetry = 10) => {
...
+        await verifierContract.deployed()
+        verifiers[circuit] = verifierContract.address
+    }
+    await new Promise((r) => setTimeout(r, DEPLOY_DELAY))
+    return verifiers
+}
+
+export const deployProofVerifiers = async (
+    deployer: ethers.Signer,
+    settings: CircuitConfig = CircuitConfig.default,
```
vivianjeng, line 75: same here

vivianjeng (changes requested):

vimwitch:
Ah yea, the verifier functions are still in `Unirep.sol`. Maybe they got put back in during a rebase in the past?

vimwitch:
@chndr1 I'm force pushing a rebase to resolve the conflicts here. I've got a backup of this branch locally.

You can pull the rebased version with

```
git checkout main
git branch -D 357-separate-proof-verification
git checkout 357-separate-proof-verification
```

vivianjeng (commented):
so the unirep users cannot use `unirep.verifyEpochKeyProof()` API anymore?
they should find or deploy a verifier to verify or decode proofs?

vimwitch:
Yes, the only proofs that are part of the unirep protocol are the UST and the default signup proof. We can supply addresses for other verifiers though.

vivianjeng:
so how can attesters use `verifyEpochKeyProof` on-chain?
and the `deployVerifiers` deploys all verifiers
which the attesters might not need them all

vivianjeng:
and if so
 all decode, verify functions in the [unirep-sol.md](https://github.com/Unirep/Unirep/blob/b35c5ed466c82d34310fafa188070f87710d6212/packages/docs/docs/contracts-api/unirep-sol.md) should be removed

chndr1:
> It would be good to also document the proof verifiers. And how to use them.

Yep, I added example code under each of the verification contract docs pages

vivianjeng:
ah I commented on the commitment and I realized that it is hard to find
I posted it here

Is there an on-chain example?
like 
```solidity
// UnirepAttester.sol
function submitAttestation(
    uint[] memory publicSignals,
    uint[8] memory proof
) public {
    // verify epoch key proof
    unirep.verifyEpochKeyProof(publicSignals, proof);
}
```

and can we take `Circuit` as one of the variables in `deployProofVerifiers()`?
so I can deploy only epochKeyVerifier with
```ts
deployProofVerifiers(accounts[0], Circuit.epochKey)
```

vimwitch:
This PR has been open for quite a long time. I think things like deploying individual proof verifiers can be done in a different PR. @vivianjeng do you think the docs are good enough to merge now? We can make more improvements after.

@chndr1 Could you rebase again? The deploy file has another conflict. I can also do it, I just don't like force pushing other peoples branches

vivianjeng (approved):
I also created another issues for this PR. #472 #473
I think the PR is a significant API change so documentation is important.
but we can merge the PR first and fix them soon

PR #430, opened 2023-04-26
Modify Synchronizer: make init _eventHandler and _eventFilter an independent function

kittybest:
## Reason ##
When developers writing their own attester, they will need to implement their own ***AppSynchronizer***, and if they emit new events in their ***UnirepApp*** contract, they need to bind the app contracts to new events. Due to the restriction of typescript, we cannot init members of the descendant class, say `class AppSynchronizer() extends Synchronizer`, before `super()` in the constructor; and right in the `super()` function, the ***Synchronizer*** already assign values to `this._eventHandlers` and `this._eventFilter`. After call `start()`, the `poll()` function will refer to these two members to decide how and what events to deal with. What we currently do in *Unirep Social* is to make another handlers and another filter, which doing the same things as the `_eventHandlers` and `_eventFilter`, I think it's not a good way for developers.
So I move the whole part of assigning `this._eventHandlers` and `this._eventFilter` to an open function, which can be called by the developer while they implementing their own attester.
btw, I think there might be better way to do this, now just try the simplest way.

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

- move initializing `this._eventHandler` and `this._eventFilter` to an independent function 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/Synchronizer.ts
+++ b/packages/core/src/Synchronizer.ts
@@ -96,7 +96,7 @@ export class Synchronizer extends EventEmitter {
             sumFieldCount: 0,
             replNonceBits: 0,
         }
-        
+
         this.assignEvents()
         this.setup().then(() => (this.setupComplete = true))
     }
--- a/packages/core/src/Synchronizer.ts
+++ b/packages/core/src/Synchronizer.ts
@@ -96,6 +96,27 @@ export class Synchronizer extends EventEmitter {
             sumFieldCount: 0,
             replNonceBits: 0,
         }
+        
+        this.assignEvents()
+        this.setup().then(() => (this.setupComplete = true))
+    }
+
+    get attesterId() {
+        if (this._attesterId.length === 0) return BigInt(0)
+        return this._attesterId[0]
+    }
+
+    get attestersOrClauses() {
+        const orClauses = [] as any[]
+        for (let id = 0; id < this._attesterId.length; id++) {
+            orClauses.push({
+                attesterId: toDecString(this._attesterId[id]),
+            })
+        }
+        return orClauses
+    }
+
+    assignEvents() {
         const allEventNames = {} as any
 
         this._eventHandlers = Object.keys(this.contracts).reduce(
@@ -168,22 +189,6 @@ export class Synchronizer extends EventEmitter {
             },
             {}
         )
-        this.setup().then(() => (this.setupComplete = true))
-    }
-
-    get attesterId() {
-        if (this._attesterId.length === 0) return BigInt(0)
-        return this._attesterId[0]
-    }
-
-    get attestersOrClauses() {
-        const orClauses = [] as any[]
-        for (let id = 0; id < this._attesterId.length; id++) {
-            orClauses.push({
-                attesterId: toDecString(this._attesterId[id]),
-            })
-        }
-        return orClauses
     }
 
     setAttesterId(attesterId: string | bigint) {


vivianjeng (commented):
I think it makes **AppSynchronizer** more complicated because the developer will modify the `assignEvents()`, which could be even worse that the private functions `this._eventHandlers` and `this._eventFilters` could be changed.
and it might not solve problem that the `assignEvents` is still in `Synchronizer` constructor, and it should be called (`super({...})`) before assigning `this.appContract` in **AppSynchronizer**

For me the best way is that

```ts
class UnirepSynchronizer {
   constructor(){
      ...
      // this.setup().then(() => (this.setupComplete = true)); // removed
   }

   async setup() {
      ...
      // set event filters, event names
      const allEventNames = {};
      ...
      this.setupComplete = true
   }
}
```

so in app synchronizer

```ts
class AppSynchronizer {
    appContract: ethers.Contract
    constructor(){
      super({...})
      this.appContract = new ethers.Contract(addr, abi, provider)
   }

    get contracts() {
        return {
            ...super.contracts,
            [this.appContract.address]: {
                contract: this.appContract,
                eventNames: [ ...]
             }
        }
    }

    async setup() {
        await super.setup()
        ...
    }
}
```

vivianjeng:
Is there a reason why `this.setup()` in the Synchronizer constructor @vimwitch ?

vimwitch:
> Is there a reason why this.setup() in the Synchronizer constructor @vimwitch ?

Yes, it's to avoid having the user make a separate call.

vimwitch:
@kittybest These variables are not designed to be overridden by subclasses. If you want to register a contract to listen for events you should override the `contracts` getter like [here](https://github.com/vimwitch/canon/blob/main/packages/relay/src/singletons/CanonSynchronizer.mjs#L13). The sychronizer will automatically look for and call event handlers as needed.

kittybest:
> @kittybest These variables are not designed to be overridden by subclasses. If you want to register a contract to listen for events you should override the `contracts` getter like [here](https://github.com/vimwitch/canon/blob/main/packages/relay/src/singletons/CanonSynchronizer.mjs#L13). The sychronizer will automatically look for and call event handlers as needed.

Sorry for late reply! @vimwitch 
I found it, and I have already modified it. But because assigning app contract **must** happened after `super` in the constructor, so while doing the setup in `super`, they won't get the contracts yet. Let me show you what I've done:
`get contracts() {
    let ret = { ...super.contracts };
    if (this.appContracts) {
      this.appContracts.map((c) => {
        ret = {
          ...ret,
          [c.address]: {
            contract: c,
            eventNames: ["SubmitDataProof"],
          },
        };
      });
    }
    return ret;
  }`
In the above code, the `appContracts` are assigned after super constructor. 
Of course I can write it like this:
`get contracts() {
    let twitter = ethers.Contract(.......)
    let ret = { ...super.contracts,
      [TWITTER_ADDRESS]: {
        contract: twitter,
         eventNames: ["SubmitDataProof"],
      }
     };
    return ret;
  }`
But that's not flexible, what if I wanna add another contract dynamically?
Also it's weird that I have to construct the contract every time I call `get`?

vimwitch:
>  Also it's weird that I have to construct the contract every time I call get?

Yes, it was designed to have the contract instance available as a variable on the class instance.

>  I found it, and I have already modified it. But because assigning app contract must happened after super in the constructor, so while doing the setup in super, they won't get the contracts yet. Let me show you what I've done:

The `setup` call in super is asynchronous and the constructor does not wait for it to complete. So any synchronous changes you make (e.g. overriding `contracts`) will happen before `setup` begins.

Can you link the code where you're running into this problem?

kittybest:
> Can you link the code where you're running into this problem?

Well, the getter `contract()` will run before `setup()`, but because I wanna pass the given attester address into the synchronizer constructor instead of write every thing inside the `contract()`. To assign given attester addresses to the synchronizer in the constructor must call `this`, but it has to be called after `super()`, so the assignment of the variable `this.appContracts` will happen after `super()`, but `setup()` is run inside `super()`, which will call the getter `contract()` before the assignment of `this.appContracts`, that's the problem I encountered. 

Also in the future, I would like to dynamically add new contracts to the synchronizer, that's why I want to make events related functions to be updatable.

See my code here: https://github.com/kittybest/my-badge/blob/relay/frontend/ranking/packages/relay/src/singletons/Synchronizer.ts

vimwitch:
Ok I see the code you posted. This shouldn't be a problem. `this.setup` is an asynchronous function so it won't be executed until after the current run loop completes. e.g. in your linked code `this.appContracts` is assigned _before_ `this.setup` executes.

In the second line of `this.setup` we use an `await`. This puts the rest of the function at the end of the run loop. The only code that is synchronously executed is `if (this.setupComplete) return`. We should add a note in `setup` saying that it shouldn't be synchronously written to make sure we don't break this functionality. But it's unlikely it will ever be synchronous because we _have_ to load data from the blockchain.

An example here showing flow of sync/async

```js
function test() {
  console.log('I log first')
  asyncLog()
  console.log('I log second')
}

async function asyncLog() {
  await new Promise(r => setTimeout(r, 0))
  console.log('I log last!')
}
```

Calling `test()` outputs the following:

```
> test()
I log first
I log second
I log last!
```

kittybest:
@vimwitch well, actually the problem is not happened in `setup()`, because binding `this.contracts` and their handlers is before `setup()` function. 
Another problem happened that there couldn't be duplicate eventName, but if I have two similar contracts, that they deployed separately but using the same contract, they will have same eventName, but it's rejected if I set up two contracts with two different addresses but with same event name.

vimwitch:
hmm yes, the system was designed for contracts that have unique event names. One approach might be changing the handler name if there are two events for two different contracts. The function name could be prefixed or suffixed with the contract address.

For the current version I'd suggest making your event names unique though.

vimwitch:
@kittybest I opened #460 to document this. I'm going to close this PR in the meantime.

vivianjeng:
Hi I think the problem is still
it is like this
https://github.com/Unirep/Unirep-Social/blob/2e1b3376399ec65f64e77fe81fef6c0227a5105e/packages/backend/src/Synchronizer.ts#L38

I have to define a contract outside the synchronizer to construct it before synchronizer config
```ts
        unirepSocialContract = new ethers.Contract(
            unirepSocialAddress,
            UNIREP_SOCIAL_ABI,
            provider
        )
        super({
            db: db ?? new MemoryConnector(constructSchema(schema)),
            unirepAddress,
            provider,
            attesterId: BigInt(unirepSocialAddress),
            prover,
        })

```

vimwitch:
I don't understand why this is necessarily a problem. The contract instance must be initialized before the `super` call, but that can happens synchronously.

vivianjeng:
I hope it can be done like this
```ts
constructor(config: {
        db?: DB
        unirepSocialAddress: string
        prover: Prover
        provider: ethers.providers.Provider
        unirepAddress: string
    }) {
        const { db, prover, unirepAddress, provider, unirepSocialAddress } =
            config
        super({
            db: db ?? new MemoryConnector(constructSchema(schema)),
            unirepAddress,
            provider,
            attesterId: BigInt(unirepSocialAddress),
            prover,
        })
        this.unirepSocialContract = new ethers.Contract(
            unirepSocialAddress,
            UNIREP_SOCIAL_ABI,
            provider
        )
    }
```
(or change the order of `super` and `this.unirepSocialContract`)
The unirep social smart contract is still deployed but I don't have to declare another variable outside the `UnirepSocialSynchronizer`

vimwitch:
Oh right you can't access `this` before `super()`. Let's queue the event handler registration asynchronously so the constructor can be used for whatever.

vimwitch:
Ok I have a pr #462 open to fix this.

vimwitch:
Closing, should be fixed in #462

Issue #356, opened 2023-03-21
Allow user to define schema for data


vimwitch:
Right now we expose an array of 253 bit integers as user data. This is hard for applications to work with because they must manually encode/decode data. Instead we should write a class for encoding and parsing data according to a schema.

The schema should allow the following types:
- `uint[1-253]`: Unsigned integers of a specific bit width (e.g. `uint3` (0-7), `uint27` (0-134217728))
- `uint`: Alias for `uint253`
- `bool`: Alias for `uint1`

Each field should specify how it's updated, `sum` or `replace`. By default the fields should be updated by replacement.

The class should allow the following:
- decode `data` array to object
- encode an update as an attestation
- print info about field structure (e.g. what fields are at what data index)

When encoding updates the library should create the minimum number of transactions.

## Example schema

```js
[
  {
    name: 'posRep',
    type: 'uint64',
    updateBy: 'sum'
  },
  {
    name: 'negRep',
    type: 'uint64',
    updateBy: 'sum'
  },
  {
    name: 'graffiti',
    type: 'uint245',
  },
  {
    name: 'postCount',
    type: 'uint49',
    updateBy: 'sum'
  },
  {
    name: 'commentCount',
    type: 'uint49',
    updateBy: 'sum'
  }
  {
    name: 'voteCount',
    type: 'uint49',
    updateBy: 'sum'
  },
  {
    name: 'averageVote',
    type: 'uint8',
  },
]
```

This schema should result in the following:
```
sumData[0]: [posRep, negRep, postCount, commentCount],
sumData[1]: [voteCount],
replaceData[0]: [graffiti, averageVote]
```

An attestation updating `posRep` and `postCount` would involve changing only 1 field.

ChialiT:
updateBy: - default value: 'replace'

chndr1:
Should be 254 bits - replacement nonce bits right?

vimwitch:
If `updateBy` is `replace` then the max type should be `253-replacementNonceBits`, yes

chndr1:
Why is `[voteCount]` in `sumData[1]` whereas `[posRep, negRep, postCount, commentCount]` are in `sumData[0]`. Additionally, would there only need to be one `sumData` field and one 'replaceData' field no matter the number of sum or replacement fields described in the schema?

PR #431, opened 2023-04-27
User-Defined Data Schemas

chndr1:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently, user data must be manually encoded and decoded. Hence, a caller requires to do some preprocessing before attesting a value.

Issue Number: #356 

## What is the new behavior?
A new `AttesterSchema` class has been implemented to use a schema to encode and decode user data. Additionally, a caller can view the location of each of the fields in the `AttesterSchema` and look up values/field indices.

- Parsing attester schemas
- Looking up schema entries by name in both the `encodedSchema` and passed `attesterSchema`
- Retrieving encoded user data and corresponding `fieldIndex`
- Handling both `sum` and `replace` updates
- Decoding the encoded data array, returning an enriched schema including user data for each schema entry

- [ ] Yes
- [x] No


packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,278 @@
...
+    constructor() {
+        this.field = []
+        this.fieldOffset = []
+        this.currentFieldSize = BigInt(0)
+
+        this.maxFieldSize = 254
+    }
+
+    public addOrPass(entry: string, size: bigint): boolean {
+        if (this.currentFieldSize + size > this.maxFieldSize) {
```
vivianjeng, line 21: if `this.maxFieldSize` is a `number` type, I think `this.currentFieldSize` and `size` don't have to be `bigint` type

packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,278 @@
...
+    private schema: {}[]
+    private schemaLookup: {}
+
+    private encodedData: bigint[]
+
+    constructor(_schema: {}[]) {
+        this.curSumDataFieldCount = 0
+        this.curReplDataFieldCount = 0
+
+        this.sumDataFields = new Array(SUM_FIELD_COUNT)
```
vivianjeng, line 94: you can do with a
```ts
this.sumDataFields = new Array(SUM_FIELD_COUNT)
                           .fill(null)
                           .map(n=> new DataField())
```

packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        for (let i = 0; i < this.replDataFields.length; i++) {
+            this.replDataFields[i] = new DataField()
+        }
+
+        this.schema = parseSchema(_schema)
+        this.encodedData = new Array(FIELD_COUNT).fill(BigInt(2) ** BigInt(254))
+
+        this.encodedData = new Array(FIELD_COUNT).fill(BigInt(0))
```
vivianjeng, line 108: is the line 106 replaced by the line 108 ?

packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,278 @@
...
+            const n = BigInt(this.replDataFields[idx].getFieldOffset(entry))
+
+            if (v >= BigInt(2) ** BigInt(bits)) {
+                throw Error('Replacement overflows uint' + bits)
+            }
+            this.encodedData[SUM_FIELD_COUNT + idx] =
+                (this.encodedData[SUM_FIELD_COUNT + idx] & ~(maxVal << n)) |
+                ((v & maxVal) << n)
+        } else {
+            throw Error('Entry does not exist in schema')
```
vivianjeng, line 236: this could be tested

packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,278 @@
...
+                }
+            }
+        } else if (updateBy == 'replace') {
+            for (let i = 0; i < FIELD_COUNT - SUM_FIELD_COUNT; i++) {
+                if (this.replDataFields[i].field.includes(entry)) {
+                    return i
+                }
+            }
+        } else {
+            throw Error('Invalid update type')
```
vivianjeng, line 205: this could be tested

packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,278 @@
...
+                    const res = this.replDataFields[
+                        this.curReplDataFieldCount
+                    ].addOrPass(name, bits)
+
+                    if (!res) {
+                        throw Error(
+                            "Couldn't replace " +
+                                name +
+                                ' replacement data field'
+                        )
```
vivianjeng, line 181: this could be tested

packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,278 @@
...
+                    !this.replDataFields[this.curReplDataFieldCount].addOrPass(
+                        name,
+                        bits
+                    )
+                ) {
+                    if (
+                        this.curReplDataFieldCount + SUM_FIELD_COUNT ==
+                        FIELD_COUNT
+                    ) {
+                        throw Error('Too many replacement and sum fields')
```
vivianjeng, line 168: this could be tested

packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,278 @@
...
+                    if (!res) {
+                        throw Error("Couldn't add " + name + ' sum data field')
+                    }
+                }
+            } else if (updateBy == 'replace') {
+                this.replDataFields[this.curReplDataFieldCount].setMaxFieldSize(
+                    254 - REPL_NONCE_BITS
+                )
+                if (bits > 254 - REPL_NONCE_BITS) {
+                    throw Error('Excessive replacement bits allocation')
```
vivianjeng, line 155: this could be tested

packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,278 @@
...
+
+                    if (this.curSumDataFieldCount >= SUM_FIELD_COUNT) {
+                        throw Error('Too many sum fields')
+                    }
+                    const res = this.sumDataFields[
+                        this.curSumDataFieldCount
+                    ].addOrPass(name, bits)
+
+                    if (!res) {
+                        throw Error("Couldn't add " + name + ' sum data field')
```
vivianjeng, line 147: this could be tested

packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,278 @@
...
+                if (
+                    !this.sumDataFields[this.curSumDataFieldCount].addOrPass(
+                        name,
+                        bits
+                    )
+                ) {
+                    this.curSumDataFieldCount++
+
+                    if (this.curSumDataFieldCount >= SUM_FIELD_COUNT) {
+                        throw Error('Too many sum fields')
```
vivianjeng, line 140: this could be tested

packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,278 @@
...
+
+        if (
+            String(entry['type']).slice(0, 4) != 'uint' ||
+            isNaN(+String(entry['type']).slice(4))
+        ) {
+            throw Error('Schema includes an invalid type')
+        }
+
+        if (entry['updateBy'] != 'sum' && entry['updateBy'] != 'replace') {
+            throw Error('Schema includes an invalid update method')
```
vivianjeng, line 73: this could be tested

packages/core/test/attesterSchema.test.ts
```diff
@@ -0,0 +1,172 @@
...
+        )
+    })
+})
+
+describe('Print Details', function () {
+    it('Print details of a schema', () => {
+        const a = new AttesterSchema(schema)
+        const schemaDetails = a.getSchemaDetails()
+
+        expect(Object.keys(schemaDetails).length).to.equal(4)
```
vivianjeng, line 102: the details could be tested

vivianjeng (commented):
good PR 👍🏻

packages/core/src/AttesterSchema.ts
chndr1: Fixed - using number now

packages/core/src/AttesterSchema.ts
chndr1: Changed. Thanks for the suggestion!

packages/core/src/AttesterSchema.ts
chndr1: I realized that the error would never be called since 
```js
const schema = this.findEntryByName(entry)
```
would throw 
```js
throw Error('Entry does not exist in schema')
```
There are now two tests that check for this, one for `update()` and the other for `getEncodedFieldIndex()`

packages/core/src/AttesterSchema.ts
chndr1: Yep, I fixed it lol

packages/core/src/AttesterSchema.ts
chndr1: Removed `Error()` for the same reasoning as for `update()`. Adding below for reference:

I realized that the error would never be called since 
```js
const schema = this.findEntryByName(entry)
```
would throw 
```js
throw Error('Entry does not exist in schema')
```
There are now two tests that check for this, one for `update()` and the other for `getEncodedFieldIndex()`

packages/core/src/AttesterSchema.ts
chndr1: Test added

packages/core/src/AttesterSchema.ts
chndr1: Reworked code so that this condition will never be met. Now, the it will repeatedly try to find available field indices until there are none left.

packages/core/src/AttesterSchema.ts
chndr1: Test added

packages/core/src/AttesterSchema.ts
chndr1: Using the same logic of repeatedly looking for available field indices. This error will no longer be reached.

vimwitch (changes requested):
This is a good first run. I think it's a bit over-engineered though. I think you don't need multiple classes, and don't need super stateful classes. This is intended to be a lightweight encoding/decoding class.

I drafted an implementation that is closer to what I'm looking for. Note that when I layout the fields in the data I'm really only looking at bit offsets. Given a bit offset you can calculate the data field index, and the offset within the field. This means there are fewer state variables laying around (which means less room for error).

I also parse the schema once, and assign a few variables that will be needed later. Once that is done parsing data can happen _very_ easily (same with encoding).

```js
export class DataSchema {
  // Make sure that the config can be a constructor arg
  // the default config might not be what is being used
  constructor(schema, config = CircuitConfig.default) {
    this.schema = this.parseSchema(schema)
    this.config = config
  }
  
  // Assign data field indices and offsets to each field
  // validate that the schema fits in the available data
  // we'll assign the following variables onto each entry in the schema
  // dataIndex: the data field index this entry is in
  // offset: the bit offset within the dataIndex that the entry _starts_ at
  // bits: the number of bits in the field, so we only have to parse the type here
  parseSchema(schema) {
    let sumOffset = 0
    let replOffset = 0
    const maxSumOffset = 253 * this.config.SUM_FIELD_COUNT
    const maxReplOffset = 253 * (this.config.FIELD_COUNT - this.config.SUM_FIELD_COUNT)
    // Create a copy of the schema that has additional variables assigned
    // on each field
    return schema.map(field => {
      const { name, type, updateBy } = field
      const bits = +type.match(/\d+$/)[0]
      if (updateBy === 'sum') {
        if (Math.floor(sumOffset / 253) !== Math.floor((sumOffset + bits) / 253)) {
          // if it doesn't fit move to the next data field index
          sumOffset += sumOffset % 253
        }
        if (sumOffset + bits >= maxSumOffset) {
          throw new Error(`Invalid schema, field "${name}" exceeds available storage`)
        }
        const dataIndex = Math.floor(sumOffset / 253)
        const offset = sumOffset % 253
        sumOffset += bits
        return { ...field, dataIndex, offset, bits }
      }
      if (updateBy === 'replace') {
        // do the same logic as sum, but use replOffset/maxReplOffset
        // make sure that dataIndex is + SUM_FIELD_COUNT
      }
      throw new Error(`Invalid updateBy strategy for field ${name}: "${updateBy}"`)
    })
  }

  // Parse some data into a JS object
  // The attester should do this often, we shouldn't store data state here
  parseData(data) {
    const parsed = {}
    // now we have a schema with fields laid out in data
    // parsing becomes much easier
    for (const field of this.schema) {
      const { name, type, updateBy, dataIndex, bits, offset } = field
      parsed[name] = (data[dataIndex] >> BigInt(offset)) & (1n << BigInt(bits))
    }
    return parsed
  }
}
```

vimwitch:
Another note, I think for replacement fields we shouldn't allow multiple subfields. e.g. if `updateBy` is `replace` the type _must_ be `uint205`. This is because we can't attest to changes without knowing the whole field. If I try to update the first 8 bits and leave the rest 0 then i've written all the other bits to 0.

So the user would have to reveal the full data so the attester could only change part of it and leave the rest the same.

In the future we could add some logic to only replace some bits to fix this problem, but for now forcing the field to be fully updated seems like a fine solution.

@vivianjeng ?

packages/core/src/AttesterSchema.ts
```diff
@@ -0,0 +1,254 @@
...
+                return Object.assign({}, this.schema[i])
+            }
+        }
+        throw Error('Entry does not exist in schema')
+    }
+
+    buildDataFields() {
+        for (let i = 0; i < this.schema.length; i++) {
+            const name = this.schema[i]['name']
+            const bits = Number(this.schema[i]['type'].slice(4))
```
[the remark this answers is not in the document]
chndr1: Regex used now

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,118 @@
...
+                replOffset += bits
+                return { ...field, dataIndex, offset, bits }
+            }
+            throw new Error(
+                `Invalid updateBy strategy for field ${name}: "${updateBy}"`
+            )
+        })
+    }
+
+    update(name: string, v: bigint): bigint {
```
vimwitch, line 85: Instead of having an `update` function to transform the data, we should have a function `buildAttestations` to apply some changes to a given data.

This is complex and we'll probably want to adjust the contracts to help with this (e.g. a multi-attest function to change multiple fields in 1 call).

So this could be something like

```ts
type Attestation = { fieldIndex: number, change: bigint }

buildAttestation(change: { name: string, v: bigint }): Attestation
buildAttestations(changes: { name: string, v: bigint }[]): Attestation[]
```

Keep in mind that if there are multiple changes to a single field they can be combined into a single attestation.

vimwitch: This is also why I don't want to track `data` as a property on this class. The attester isn't meant to know the user data, and will often attest without knowing the full user data. The attester will also receive partial proofs of data. e.g. a user might prove bits 32-64 of field index 2 to the attester and leave all other fields as 0. There's not much use in storing that on the instance of the class. It's better to think of this class like the [ethers abi coder](https://docs.ethers.org/v6/api/abi/abi-coder/).

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,118 @@
+import { CircuitConfig } from '@unirep/circuits'
+
+export class DataSchema {
+    public schema: any[]
```
vimwitch, line 4: The schema can be typed, something like

```ts
type SchemaField = {
  name: string,
  type: string,
  updateBy: 'sum' | 'replace'
}

public schema: SchemaField[]
```

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    parseSchema(schema: any[]): any[] {
+        let sumOffset = 0
+        let replOffset = 0
+        const maxSumOffset = 253 * this.config.SUM_FIELD_COUNT
+        const maxReplOffset =
+            253 * (this.config.FIELD_COUNT - this.config.SUM_FIELD_COUNT)
+
+        return schema.map((field, idx) => {
+            const { name, type, updateBy, ...extranousFields } = field
```
vimwitch, line 22: "extraneous", this could also be called "extra"

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,118 @@
...
+                    )
+                }
+                const dataIndex = Math.floor(sumOffset / 253)
+                const offset = sumOffset % 253
+                sumOffset += bits
+                return { ...field, dataIndex, offset, bits }
+            } else if (updateBy === 'replace') {
+                if (bits !== 253 - this.config.REPL_NONCE_BITS)
+                    throw new Error(
+                        `Invalid uint size for field ${name}: ${bits}`
```
vimwitch, line 65: Maybe in the error message add a note saying `field must be ${253 - this.config.REPL_NONCE_BITS} bits`

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        this.data[idx] =
+            (this.data[idx] & ~(maxVal << BigInt(field.offset))) |
+            ((v & maxVal) << BigInt(field.offset))
+
+        return this.data[idx]
+    }
+
+    parseData(data: bigint[]): {} {
```
vimwitch, line 105: Specify the return value as `any` instead of `{}`

vimwitch (changes requested):

vivianjeng:
> In the future we could add some logic to only replace some bits to fix this problem,

sounds good

we should change the circuits and add the logic to data schema

packages/core/src/DataSchema.ts
chndr1: Done, returning `any` now

packages/core/src/DataSchema.ts
chndr1: Done, using this error message now

packages/core/src/DataSchema.ts
chndr1: Done, using `extraFields` now

packages/core/src/DataSchema.ts
chndr1: Done, using `SchemaField` now

packages/core/src/DataSchema.ts
chndr1: Done, `buildAttestation` and `buildAttestations` now included. `buildAttestation` combines taking all changes that correspond to the same `fieldIndex` and altering `change` value of the attestation going to that field. Additionally, multiple `sum` attestations are combined, where the additions are performed in `buildAttestations` with error handling. Both `buildAttestation` and `buildAttestations` return an `Attestation {fieldIndex, change}` type struct. `fieldIndex` is found through a lookup of the data schema created in `parseSchema`

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,173 @@
...
+                `Invalid updateBy strategy for field ${name}: "${updateBy}"`
+            )
+        })
+    }
+
+    buildAttestation(change: { name: string; v: bigint }): Attestation {
+        const field: any = this.schema.find((f) => f.name === change.name)
+
+        if (field === undefined) {
+            throw new Error(`${change.name} not found`)
```
vimwitch, line 99: Can be more explicit like `Field "${change.name}" not found`

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,173 @@
...
+                replOffset += bits
+                return { ...field, dataIndex, offset, bits }
+            }
+            throw new Error(
+                `Invalid updateBy strategy for field ${name}: "${updateBy}"`
+            )
+        })
+    }
+
+    buildAttestation(change: { name: string; v: bigint }): Attestation {
```
vimwitch, line 95: Prefer more verbose names for variables. So instead of `v` maybe `value` or `change`

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,173 @@
...
+
+        const attestation: Attestation = {
+            fieldIndex,
+            change: x,
+        }
+
+        return attestation
+    }
+
+    buildAttestations(changes: { name: string; v: bigint }[]): Attestation[] {
```
vimwitch, line 118: Input variable name here

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,173 @@
...
+        let attestations: Attestation[] = Array(this.schema.length).fill(null)
```
vimwitch, line 119: This can be a `const`

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            const fieldIndex: number = field.dataIndex
+            let v: bigint = change.v << BigInt(field.offset)
+            const maxVal: bigint = (BigInt(1) << BigInt(field.bits)) - BigInt(1)
+
+            if (attestations[fieldIndex] !== null) {
+                v =
+                    change.v +
+                    (field.updateBy === 'sum'
```
vimwitch, line 134: This ternary should be broken into `if` statements. It's hard to read here.

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,173 @@
...
+                        ? (attestations[fieldIndex].change >>
+                              BigInt(field.offset)) &
```
vimwitch, line 136: Instead of silently taking the lower bits with `&` we should throw an error if an attestation overflows the field bits.

packages/core/src/DataSchema.ts
```diff
@@ -0,0 +1,173 @@
...
+                          maxVal
+                        : BigInt(0))
+            }
+
+            if (v > maxVal << BigInt(field.offset)) {
+                throw new Error(`${change.name} exceeds allocated space`)
+            }
+
+            if (attestations[fieldIndex] !== null)
+                v =
```
vimwitch, line 146: This is very complicated, you can just do `v = v + attestations[fieldIndex].change`

vimwitch (changes requested):

packages/core/src/DataSchema.ts
vimwitch: Unless it's a replacement field in which case it should be no operation (discard the previous attestations)

packages/core/src/DataSchema.ts
chndr1: `value` is preserved and `change` is redundant with the "change" dictionary, so I'm renaming them to `val`

packages/core/src/DataSchema.ts
chndr1: This is meant to select the existing `change` value stored in `attestations[]` from previous attestations, so this isn't supposed to be an overflow check. Simply a way to get the current `change` value to be updated.

chndr1: Now that I think about it, we don't need to extract the existing value and overwrite the attestation `change`, so this is an unnecessary step

chndr1: Also, an error is thrown if an attestation overflows the field bits:
```throw new Error(`${change.name} exceeds allocated space`)```

File: packages/core/src/DataSchema.ts (hunk 25 above)
Comment (chndr1) [reply packages/core/src/DataSchema.ts:146]: Yep, I realize I was making it a bit more complicated than necessary by trying to overwrite the existing change value with bit shifts instead of just adding to it.

File: packages/core/src/DataSchema.ts (hunk 24 above)
Comment (vimwitch) [reply packages/core/src/DataSchema.ts:136]: Oh okay. This isn't necessary, you can combine the old changes by doing simple addition.

Comment (vimwitch) [reply packages/core/src/DataSchema.ts:136]: Like you can combine the entire field with the shifted change value

File: packages/core/src/DataSchema.ts [hunk 26]
```diff
@@ -0,0 +1,173 @@
...
+    parseSchema(schema: SchemaField[]): any[] {
+        let sumOffset = 0
+        let replOffset = 0
+        const maxSumOffset = 253 * this.config.SUM_FIELD_COUNT
+        const maxReplOffset =
+            253 * (this.config.FIELD_COUNT - this.config.SUM_FIELD_COUNT)
+
+        return schema.map((field, idx) => {
+            const { name, type, updateBy, ...extraFields } = field
+            const match = type.match(/^uint(\d+)$/)
```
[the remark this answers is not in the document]
chndr1: As discussed, `bool` logic will be postponed until a later version since we can't replace a fractional component of a replacement field

packages/core/src/DataSchema.ts
chndr1: Change added,

packages/core/src/DataSchema.ts
chndr1: `v` is now `val`

packages/core/src/DataSchema.ts
chndr1: Ternary removed in favor of a simple if statement

packages/core/src/DataSchema.ts
chndr1: To summarize, the part of the code that figures out whether there is an attestation overflow or not is the following:
```js
            // Get existing attestation sum value
            let prevVal: bigint = BigInt(0)

            if (field.updateBy === 'sum' && attestations[fieldIndex] !== null)
                prevVal =
                    (attestations[fieldIndex].change >> BigInt(field.offset)) &
                    maxVal

            if (v + prevVal > maxVal << BigInt(field.offset)) {
                throw new Error(`${change.name} exceeds allocated space`)
            }

```
In this, I first get the previous value stored in the attestation using bitwise operations. Then, I add this to the current change value and check if that exceeds the allocated space, which follows the overflow check you suggested.

vimwitch (commented):
I think this LGTM, just needs some docs

packages/docs/docs/core-api/data-schema.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Data Schema 
+---
+
+User-defined data schemas can be used when interfacing with the exposed `uint253` attestation data. This tool encodes and decodes attestation data, consolidating a user-defined attestation changes.
```
vimwitch, line 5: This entry section needs some example code showing importing the module, an example schema, and using the functions containing in here.

vimwitch, line 5: I'd replace this entry text with "The `DataSchema` class abstracts unirep data into a javascript object. This class can be used to encode and decode attestation data, and build attestations that are ready to be submitted to the unirep smart contract.

chndr1: Example added

chndr1: Entry text added

packages/docs/docs/core-api/data-schema.md
```diff
@@ -2,15 +2,100 @@
...
+    {
+        name: 'voteCount',
+        type: 'uint49',
+        updateBy: 'sum',
+    },
+]
+
+const d = new DataSchema(schema)
+
+// The schema is parsed in the `DataSchema` constructor, but can also be manually parsed
```
vimwitch, line 45: I think this part is unnecessary, I imagined `parseSchema` being a private function

chndr1: Deleted

vimwitch (approved):

PR #433, opened 2023-05-02
feat(core): remove prover from synchronizer

### LINKED ISSUE #336: Remove `prover` from Synchronizer
The `Synchronizer` doesn't generate proofs. Instead the `UserState` should accept the prover.

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: close #336 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


packages/core/src/UserState.ts
```diff
@@ -38,26 +39,32 @@ export default class UserState {
...
+            prover: Prover
+        }
     ) {
+        const { id, prover } = userConfig
+        if (!id) {
+            throw new Error(
+                '@unirep/core:UserState: id must be supplied as an argument when initialized'
+            )
+        }
+        if (!prover) {
```
vimwitch, line 56: This constructor is now a bit of a mess. I think we should look for either `userConfig.prover` or `config.prover` to maintain backward compatibility.

vimwitch (changes requested):

packages/core/src/UserState.ts
vimwitch: The user might supply a config as the first argument (not necessarily a synchronizer). The prover could be in this argument.

If we're going to break backward compatibility we should make this a bit more clean?

Maybe like

```ts
constructor(config: {
  synchronizer?: Synchronizer,
  db?: DB,
  attesterId?: bigint | string,
  unirepAddress?: bigint | string,
  provider?: ethers.providers.Provider,
  id: Identity,
  prover: Prover
}) {
  if (config.synchronizer) {
    this.sync = config.synchronizer
  } else {
    this.sync = new Synchronizer(config)
  }
  this.id = config.id
  this.prover = config.prover
}
```

packages/core/src/UserState.ts
```diff
@@ -25,39 +25,59 @@ import { Synchronizer, toDecString } from './Synchronizer'
...
+                '@unirep/core:UserState: prover must be supplied as second argument when initialized with a sync'
+            )
+        }
+        if (synchronizer) {
+            this.sync = synchronizer
+        } else {
+            if (!provider || !unirepAddress) {
                 throw new Error(
-                    '@unirep/core:UserState: id must be supplied as second argument when initialized with a sync'
+                    '@unirep/core:UserState: provider and Unirep address must be supplied as second argument when initialized with a sync'
```
vimwitch, line 69: This error message should not have the `supplied as second argument` text. There is only 1 argument now

packages/core/src/UserState.ts
```diff
@@ -25,39 +25,59 @@ import { Synchronizer, toDecString } from './Synchronizer'
...
+            db,
+        } = config
+        if (!id) {
+            throw new Error(
+                '@unirep/core:UserState: id must be supplied as second argument when initialized with a sync'
+            )
+        }
+        if (!prover) {
+            throw new Error(
+                '@unirep/core:UserState: prover must be supplied as second argument when initialized with a sync'
```
vimwitch, line 61: Same here, just say it needs to be supplied

packages/core/src/UserState.ts
```diff
@@ -25,39 +25,59 @@ import { Synchronizer, toDecString } from './Synchronizer'
...
+            synchronizer,
+            attesterId,
+            unirepAddress,
+            provider,
+            prover,
+            db,
+        } = config
+        if (!id) {
+            throw new Error(
+                '@unirep/core:UserState: id must be supplied as second argument when initialized with a sync'
```
vimwitch, line 56: Same here

vimwitch (changes requested):
Looks good, error messages need to be updated. Docs should be updated too

vimwitch (approved):

PR #459, opened 2023-05-15
fix(contracts): change EpochKeyData layout

### LINKED ISSUE #453: [Audit warning] Wasted Storage
The Unirep protocol allows Attesters to declare data that is unique to individual epoch keys and users. To hold information about this data, the `IUnirep` interface declares the `EpochKeyData` struct which allocates a static array of size 30 to hold the data. The `Unirep` contract, however, uses the immutable variable `fieldCount` as the size of the attester data without comparing it against 30.

> Unnecessarily large `data` struct field
> 
> 
> ```ts
>     struct EpochKeyData {
>         uint256 leaf;
>         uint256[30] data;
>         uint48 leafIndex;
>         uint48 epoch;
>     }
> ```
> 

## Impact

In the event that `fieldCount` is less than 30, storage will be wasted as the `Unirep` contract maintains an `EpochKeyData` entry for every epoch key that is attested to. If `fieldCount` is greater than 30, the contract will be broken.

## Recommendation

Ideally change the size of `data` to be `fieldCount`. If that is not done, add a requirement that `fieldCount <= 30` in the constructor to prevent potential initialization errors.

vimwitch:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Closes #453 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/contracts/contracts/Unirep.sol
+++ b/packages/contracts/contracts/Unirep.sol
@@ -52,6 +52,8 @@ contract Unirep is IUnirep, VerifySignature {
         IVerifier _epochKeyVerifier,
         IVerifier _epochKeyLiteVerifier
     ) {
+        // see IUnirep.sol EpochKeyData.data
+        require(_config.fieldCount < 128, 'datasize');
         stateTreeDepth = _config.stateTreeDepth;
         epochTreeDepth = _config.epochTreeDepth;
         historyTreeDepth = _config.historyTreeDepth;
--- a/packages/circuits/src/CircuitConfig.ts
+++ b/packages/circuits/src/CircuitConfig.ts
@@ -25,6 +25,18 @@ export class CircuitConfig {
         return new CircuitConfig(defaultConfig)
     }
 
+    get contractConfig() {
+        return {
+            stateTreeDepth: this.STATE_TREE_DEPTH,
+            epochTreeDepth: this.EPOCH_TREE_DEPTH,
+            historyTreeDepth: this.HISTORY_TREE_DEPTH,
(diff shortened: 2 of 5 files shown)


packages/contracts/contracts/Unirep.sol
```diff
@@ -52,6 +52,8 @@ contract Unirep is IUnirep, VerifySignature {
         IVerifier _epochKeyVerifier,
         IVerifier _epochKeyLiteVerifier
     ) {
+        // see IUnirep.sol EpochKeyData.data
+        require(_config.fieldCount < 128, 'datasize');
```
vivianjeng, line 56: I am not sure if this should be reverted by an error (to be consistent with the following codes)

but I found that in our deploy scripts, the error cannot be shown because we retry if it fails ( so it would be difficult to detect if the transaction fails because of which reason)

vivianjeng: it also reminds me that we should have tests for deploy scripts
we just assume it succeed in current tests

vimwitch: yea, i think we don't need an error code for this. In the deploy script we should check the `config` values before trying to send the deploy transaction.

vivianjeng (commented):
so is a deployment test going to be included?

vivianjeng (commented):
I think there could be a error message in deploy script to say the data size should be less than 128 (if the size is set more than 128)

vivianjeng (approved):

PR #464, opened 2023-05-17
Fix manual signup

### LINKED ISSUE #455: [Audit low severity] Manual Signup may be Incorrect
The Unirep protocol gives attesters the ability to specify unique initial data if desired by using the `manualUserSignUp` function. To do so, the attester must specify the state tree leaf along with initial data associated with the account. The state tree leaf, however, is the hash of several pieces of information, including the initial account data. Currently there is no guarantee, however, that the data provided to `manualUserSignUp` is the same data that was used to calculate the `stateTreeLeaf`.

> Location where an attester can specify alternate initial data
> 
> 
> ```solidity
>     function manualUserSignUp(
>         uint48 epoch,
>         uint256 identityCommitment,
>         uint256 stateTreeLeaf,
>         uint256[] calldata initialData
>     ) public {
>         _userSignUp(epoch, identityCommitment, stateTreeLeaf);
>         if (initialData.length > fieldCount) revert OutOfRange();
>         for (uint8 x = 0; x < initialData.length; x++) {
>             if (initialData[x] >= SNARK_SCALAR_FIELD) revert InvalidField();
>             if (
>                 x >= sumFieldCount &&
>                 initialData[x] >= 2 ** (254 - replNonceBits)
>             ) revert OutOfRange();
>             emit Attestation(
>                 type(uint48).max,
>                 identityCommitment,
>                 uint160(msg.sender),
>                 x,
>                 initialData[x]
>             );
>         }
>     }
> ```
> 

## Impact

If the data used to calculate the `stateTreeLeaf` is inconsistent with `initalData`, a user will not be able to use the protocol as knowledge of the data is required to allow a user to properly transition. Therefore, an attester can accidentally or maliciously prevent a user from using their account. Note, that such a user could always sign up with a separate identify commitment but it could be an inconvenience (for example, if there were a reason to use the same `identityCommitment` between applications). 

## Recommendation

While we recognize that the current calculation of the `stateTreeLeaf` would prevent such validation, the leaf calculation could be changed to `h(h(data), h(identity_secret, attester_id, epoch))`.

vivianjeng:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: fix #455 

## What is the new behavior?

The previous `manualUserSignUp` it only takes `stateTreeLeaf` but it doesn't check the relation between `stateTreeLeaf` and `initialData`.
The proposed solution is that
in circuits: all initial data would be the private input, and output the `data_hash`
and the original `userSignUp` will check if `data_hash` is computed by all 0 data
and the `manualUserSignUp` will also check if `data_hash` matches hash of `initialData`

There is also an optimization that we can encode `attesterId` and `epoch` in one public signal.
So currently there are 4 public signals like before:  `[idCommitment, stateTreeLeaf, dataHash, control]`
(where `control = attesterId + 2**160 * epoch`)

- [x] Yes
- [ ] No


vivianjeng:
@vimwitch when I changed the signup circuit, the web prover will fail 😅

packages/circuits/circuits/leafHasher.circom
```diff
@@ -61,10 +63,13 @@ template StateTreeLeaf(FIELD_COUNT) {
...
 
+  control <== attester_id + 2**160*epoch;
+  data_hash <== hasher[FIELD_COUNT-2].out;
+
   component final_hasher = Poseidon(3);
   final_hasher.inputs[0] <== identity_secret;
-  final_hasher.inputs[1] <== attester_id + 2**160*epoch;
-  final_hasher.inputs[2] <== hasher[FIELD_COUNT-2].out;
+  final_hasher.inputs[1] <== control;
+  final_hasher.inputs[2] <== data_hash;
```
vimwitch, line 72: I think we should adjust the state tree leaf so it's like

```
H(H(identitySecret, epoch, attesterId), H(data))
```

We can call the `H(identitySecret, epoch, attesterId)` the state leaf identity hash, and `H(data)` the state leaf data hash.

This way an application can provide the state leaf identity hash, and calculate the data hash onchain to build the state leaf.

packages/contracts/contracts/Unirep.sol
```diff
@@ -90,27 +98,36 @@ contract Unirep is IUnirep, VerifySignature {
      * e.g. to insert a non-zero data field in the state tree leaf
      **/
     function manualUserSignUp(
-        uint48 epoch,
-        uint256 identityCommitment,
-        uint256 stateTreeLeaf,
+        uint256[] calldata publicSignals,
```
vimwitch, line 101: I think this function should not accept a proof. The idea with this is to allow attesters to define their own signup proof that is verified in their own contract. We can accept the `stateLeafIdentityHash`, and `data` then calculate the state tree leaf like `H(stateLeafIdentityHash, H(data))`.

packages/contracts/contracts/Unirep.sol
```diff
@@ -121,27 +138,24 @@ contract Unirep is IUnirep, VerifySignature {
         uint256[] calldata publicSignals,
         uint256[8] calldata proof
     ) public {
-        uint256 attesterId = publicSignals[2];
+        SignupSignals memory signals = decodeSignupSignals(publicSignals);
         // only allow attester to sign up users
-        if (uint256(uint160(msg.sender)) != attesterId)
-            revert AttesterIdNotMatch(uint160(msg.sender));
+        if (signals.dataHash != defaultUserDataHash) revert InvalidDataHash();
```
vimwitch, line 143: I think this proof shouldn't include a `dataHash` value. Instead it can output the `stateLeafIdentityHash`. This function then can be overloaded like

```solidity
function userSignUp(
  uint[] calldata publicSignals,
  uint[8] calldata proof
) {}

function userSignUp(
  uint[] calldata publicSignals,
  uint[8] calldata proof,
  uint[] calldata data
) {}
```

If there is no data we calculate the state leaf as `H(stateLeafIdentityHash, defaultUserDataHash)`. If there is data we do it as `H(stateLeafIdentityHash, H(data))`.

Only the `stateLeafIdentityHash` _must_ be calculated in ZK because it has a secret value.

vimwitch (changes requested):

packages/contracts/contracts/interfaces/IUnirep.sol
```diff
@@ -74,6 +75,13 @@ interface IUnirep {
     error InvalidHistoryTreeRoot(uint256 historyTreeRoot);
     error InvalidStateTreeRoot(uint256 stateTreeRoot);
 
+    struct SignupSignals {
+        uint256 stateTreeLeaf;
+        uint256 epoch;
```
vimwitch, line 80: This can be `uint48` and `attesterId` can be `uint160`

packages/circuits/test/signup.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        )
+        expect(data.control.toString()).to.equal(
+            SignupProof.buildControl({ attesterId, epoch }).toString()
+        )
+    })
+
+    it('should fail to prove an out of range attesterId', async () => {
+        const id = new Identity()
+        const epoch = 0
+        const attesterId = BigInt(1) >> BigInt(160)
```
vimwitch, line 82: This is being shifted the wrong direction, it results in `attesterId == 0`

packages/circuits/test/signup.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        await new Promise<void>((rs, rj) => {
+            genProofAndVerify(Circuit.epochKeyLite, circuitInputs)
+                .then(() => rj())
+                .catch(() => rs())
+        })
+    })
+
+    it('should fail to prove an out of range epoch', async () => {
+        const id = new Identity()
+        const epoch = 1 >> 48
```
vimwitch, line 97: Same here, this results in `0`. This should also be a `bigint`, js starts having trouble with numbers near 2^50

packages/contracts/contracts/Unirep.sol
```diff
@@ -94,17 +94,23 @@ contract Unirep is IUnirep, VerifySignature {
     function manualUserSignUp(
         uint48 epoch,
         uint256 identityCommitment,
-        uint256 stateTreeLeaf,
+        uint256 leafIdentityHash,
         uint256[] calldata initialData
     ) public {
-        _userSignUp(epoch, identityCommitment, stateTreeLeaf);
+        if (initialData.length == 0) revert ZeroInitialData();
```
vimwitch, line 100: I think we should allow `0` data signups. We should use the pre-calculated 0 data hash for this.

Maybe overload this function with 

```solidity
    function manualUserSignUp(
        uint48 epoch,
        uint256 identityCommitment,
        uint256 leafIdentityHash
    ) public
```

vimwitch (changes requested):

packages/contracts/contracts/Unirep.sol
vivianjeng: I prefer not to overload the function or it will be called like `unirepContract['manualUserSignUp(uint48,uint256,uint256)']`

vimwitch (approved):
nice, lgtm now, just needs to be rebased

PR #470, opened 2023-05-24
wip: initial commit for util contract

yuriko627:
## Pull request checklist

- [ ] Build (`yarn build`) was run locally in root directory and any changes were pushed
- [ ] Lint (`yarn lint:fix`) was run locally in root directory and any changes were pushed

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
+++ b/packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
@@ -1,0 +1,114 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { Unirep } from "../Unirep.sol";
+
+/**
+ * @title ReimburseAttestation
+ * @dev An attester can inherit this contract to hold budget and allow the callers of the contract (attester app users) to be reimbersed for attestations.
+ 	This would make the publicly funding operation of attesters more modular and transparent.
+	e.g. An attester could directly get funded from a gitcoin round for its operation.
+ */
+contract ReimburseAttestation {
+
+	uint public fund;
+    bool public accpetDonations;
+
+	mapping(address => bool) public whitelist;
+
+    Unirep public unirep;
+
+    // id of the attester this budget contract is deployed for
+    uint256 immutable public attesterId;
+
+    event Reimbursed(address indexed recepient, uint amount, string txType);
+
+    constructor (
+        Unirep _unirep,
+        uint256 _attesterId
+    ) {
+        unirep = _unirep;
+        attesterId = _attesterId;
+        accpetDonations = true;
+    }
+
+    /**
+     * @dev Add users to whitelist
+     */
+    function addToWhitelist(address[] calldata addresses) external {
+        for (uint i = 0; i < addresses.length; i++) {
+            whitelist[addresses[i]] = true;
+        }
+    }
+
+    /**
+     * @dev Remove users from whitelist
+     */
+    function removeFromWhitelist(address[] calldata toRemoveAddresses) external {
+        for (uint i = 0; i < toRemoveAddresses.length; i++) {
+            delete whitelist[toRemoveAddresses[i]];
+        }
+    }
+
+    /**
+     * @dev The contract should hold the funds and accept funds from anyone.
+     */
+    function donate() public payable {
+        fund += msg.value;
+    }
+
+    /**
+     * @dev Restrict/allow the contract to receive more donations.
+
+     */
+    function toggleAcceptingDonations() public {
+        require(msg.sender == address(this), "Only this budget contract itself can call this function");
+        accpetDonations = !accpetDonations;
+    }
+
+    function userStateTransition(uint[] calldata publicSignals, uint[8] calldata proof) public {
+        require(whitelist[msg.sender], "This user is not in the whitelist");
+
+        uint attesterSignalIndex = 3 + unirep.numEpochKeyNoncePerEpoch();
+
+        // check that we're doing a UST for the attester this budget contract is deployed for
+        require(publicSignals[attesterSignalIndex] == attesterId);
+
+        // start measuring gas
+        uint startGas = gasleft();
+
+        // this is guaranteed not to re-enter if it's configured correctly
+        unirep.userStateTransition(publicSignals, proof);
+
+        // using basefee to prevent large gasprice values
+        uint txCost = (startGas - gasleft()) * block.basefee;
+
+        payable(msg.sender).transfer(txCost);
+
+        emit Reimbursed(msg.sender, txCost, 'UST');
+    }
+
+	function userSignUp(uint[] calldata publicSignals, uint[8] calldata proof) public payable {
+        require(whitelist[msg.sender], "This user is not in the whitelist");
+
+        uint attesterSignalIndex = 3 + unirep.numEpochKeyNoncePerEpoch();
+
+        // check that we're doing a userSignUP for the attester this budget contract is deployed for
+        require(publicSignals[attesterSignalIndex] == attesterId);
+
+        // start measuring gas
+        uint startGas = gasleft();
+
+        // this is guaranteed not to re-enter if it's configured correctly
+        unirep.userSignUp(publicSignals, proof);
+
+        // use BASEFEE here to prevent large gasprice values
+        uint txCost = (startGas - gasleft()) * block.basefee;
+
+        payable(msg.sender).transfer(txCost);
+
+        emit Reimbursed(msg.sender, txCost, 'userSignUp');
+
+    }
+
+}


vivianjeng (changes requested):
There are some debug messages e.g. `console.log` and comment out lines
please remove them if they is no need for them

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,150 @@
...
+import 'hardhat/console.sol';
+
+/**
+ * @title ReimburseAttestation
+ * @dev An attester can inherit this contract to hold budget and allow the callers of the contract (attester app users) to be reimbersed for attestations.
+ 	This would make the publicly funding operation of attesters more modular and transparent.
+	e.g. An attester could directly get funded from a gitcoin round for its operation.
+ */
+contract ReimburseAttestation {
+    bool public accpetDonations;
```
vimwitch, line 14: There's a typo here

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    mapping(address => bool) public whitelist;
+
+    Unirep public unirep;
+
+    // id of the attester this budget contract is deployed for
+    uint256 public attesterId;
+
+    event Reimbursed(address indexed recepient, uint amount, string txType);
```
vimwitch, line 23: typo here too, `recipient`

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,150 @@
...
+ */
+contract ReimburseAttestation {
+    bool public accpetDonations;
+
+    mapping(address => bool) public whitelist;
+
+    Unirep public unirep;
+
+    // id of the attester this budget contract is deployed for
+    uint256 public attesterId;
```
vimwitch, line 21: This should be a `uint160` or an `address`

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,150 @@
...
+        emit FundsReceived(msg.sender, msg.value);
+    }
+
+    /**
+     * @dev Restrict/allow the contract to receive more donations.
+
+     */
+    function toggleAcceptingDonations() public {
+        require(
+            msg.sender == address(this),
```
vimwitch, line 83: I think this is condition is impossible? Maybe allow the address `attesterId` to control this, or add an `owner` variable?

packages/contracts/test/reimburseAttestation.test.ts
```diff
@@ -0,0 +1,126 @@
+//@ts-ignore
+import { ethers } from 'hardhat'
+import { expect } from 'chai'
+import { deployUnirep } from '@unirep/contracts/deploy'
+import { Identity } from '@semaphore-protocol/identity'
+import { schema, UserState } from '@unirep/core'
```
vimwitch, line 6: We try to avoid depending on the `@unirep/core` package in the contracts package. It creates a circular dependency. Instead what you could/should do here is not use a _real_ unirep instance, but instead deploy a contract like

```solidity
contract FakeUnirep {
  function userSignUp(uint[] calldata publicSignals, uint[8] calldata proof) public {}
  function userStateTransition(uint[] calldata publicSignals, uint[8] calldata proof) public {}
}
```

This way you don't need to generate a valid ZK proof, you can just test the reimburse contract logic.

vimwitch: There should probably be _some_ cases that are tested with real zk proofs just to make sure that the `attesterId` variable is being extracted correctly (for example). But it's okay to omit it for now because the UST public signals are all pseudo-random, so if the `attesterId` public signal is incorrect the contract simply won't work, it won't be exploitable.

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,150 @@
...
+            'Only this budget contract itself can call this function'
+        );
+        accpetDonations = !accpetDonations;
+    }
+
+    function userSignUp(
```
vimwitch, line 89: It would be good to add an option to enable user sign up re-reimbursement. Only re-reimbursing USTs is probably a safer approach to start.

vimwitch (changes requested):
structure looks good, tests can be simplified

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,150 @@
...
+        uint256[] calldata publicSignals,
+        uint[8] calldata proof
+    ) public payable virtual {
+        require(whitelist[msg.sender], 'This user is not in the whitelist');
+
+        uint256 attesterSignalIndex = unirep.numEpochKeyNoncePerEpoch();
```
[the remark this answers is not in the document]
yuriko627: so is it okay to just hardcode like `uint256 attesterSignalIndex = 2;`?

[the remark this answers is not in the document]
vimwitch: I think yes. In the future we could define an abstract contract that defines the signal indices as constants.

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,150 @@
...
+        emit Reimbursed(msg.sender, txCost, 'userSignUp');
+    }
+
+    function userStateTransition(
+        uint[] calldata publicSignals,
+        uint[8] calldata proof
+    ) public payable virtual {
+        require(whitelist[msg.sender], 'This user is not in the whitelist');
+
+        uint attesterSignalIndex = unirep.numEpochKeyNoncePerEpoch();
```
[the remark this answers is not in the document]
vimwitch: In the unirep contract we have `uint256 attesterSignalIndex = 3 + numEpochKeyNoncePerEpoch;`.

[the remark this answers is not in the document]
yuriko627: For consistency, I'd just hardcode for this as well for now. Is that okay?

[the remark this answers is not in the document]
vimwitch: Yes, if you mean add 3 to `unirep.numEpochKeyNoncePerEpoch()`

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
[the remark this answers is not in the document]
yuriko627: so should this be `uint256 attesterSignalIndex = unirep.numEpochKeyNoncePerEpoch()` to be consistent with what you suggest for the UST function?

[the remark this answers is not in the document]
vimwitch: No, for the UST there is a variable number of signals based on the configured `numEpochKeyNoncePerEpoch`. For the signup proof there is a constant number of signals. So you can simply use the value `2`

packages/contracts/test/reimburseAttestation.test.ts
```diff
@@ -0,0 +1,126 @@
...
+        await app.addToWhitelist([owner], owner)
+        const tx = await app.userSignUp(publicSignals, proof)
+        const receipt = await tx.wait()
+
+        expect(receipt.status).equal(1)
+        // TODO: find something equivalent to numUserSignUPs in unirep contract
+        // const numUserSignUPs = await unirep.numUserSignUPs
+        // expect(1).equal(numUserSignUPs)
+        userState.sync.stop()
+    })
```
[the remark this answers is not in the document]
yuriko627: I added an assertion to check the caller gets the correct amount of reimbursement, but the value never matches. I've been trying to fix it but can't tell what's wrong. I logged some values, and it seems like the balance of the caller reduces after the transaction. Any idea? (I pushed the updates, so you can see the source code in my latest commit)

packages/contracts/test/reimburseAttestation.test.ts
yuriko627: I've added `MockUnirep.sol` and trying to replicate the same behavior but get this error for UST function call. Do you have any idea why? Function name and signature should be correct
```
 Error: Transaction reverted: function selector was not recognized and there's no fallback function
    at MockUnirep.<unrecognized-selector> (contracts/test/MockUnirep.sol:9)

Comment (yuriko627) [reply packages/contracts/test/reimburseAttestation.test.ts]: You can see what I added in my latest commit

File: packages/contracts/test/reimburseAttestation.test.ts (hunk 9 above)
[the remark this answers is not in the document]
Comment (yuriko627) [reply packages/contracts/test/reimburseAttestation.test.ts]: resolved this :)

File: packages/contracts/test/reimburseAttestation.test.ts (hunk 5 above)
Comment (yuriko627) [reply packages/contracts/test/reimburseAttestation.test.ts]: resolved this too :)

File: packages/contracts/contracts/attester-utils/ReimburseAttestation.sol [hunk 10]
```diff
@@ -0,0 +1,154 @@
...
+    Unirep public unirep;
+
+    // id of the attester this budget contract is deployed for
+    uint160 public attesterId;
+
+    enum TxType {
+        Signup,
+        UST
+    }
+    TxType public txType;
```
vivianjeng, line 26: It doesn't have to be a public variable. When you emit events, you just pass a `TxType.Signup` or `TxType. UST`.

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,154 @@
...
+            address(this).balance >= reimbursement,
+            'Not enough budget in the contract'
+        );
+
+        (bool sent, ) = payable(msg.sender).call{value: reimbursement}('');
+        // The gas cost will be deducted from the external account that initiated the transaction.
+
+        require(sent, 'Failed to send Ether');
+
+        txType = TxType.Signup;
```
vivianjeng, line 123: can be removed

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,154 @@
...
+
+        // this is guaranteed not to re-enter if it's configured correctly
+        unirep.userStateTransition(publicSignals, proof);
+
+        // using basefee to prevent large gasprice values
+        uint txCost = (startGas - gasleft()) * block.basefee;
+
+        payable(msg.sender).transfer(txCost);
+
+        txType = TxType.UST;
```
vivianjeng, line 150: can be removed

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,154 @@
...
+pragma solidity ^0.8.0;
+import {Unirep} from '../Unirep.sol';
+
+/**
+ * @title ReimburseAttestation
+ * @dev An attester can inherit this contract to hold budget and allow the callers of the contract (attester app users) to be reimbersed for attestations.
+ 	This would make the publicly funding operation of attesters more modular and transparent.
+	e.g. An attester could directly get funded from a gitcoin round for its operation.
+ */
+contract ReimburseAttestation {
```
vivianjeng, line 11: I think it is not an `attestation` here.
It is a more general transaction (e.g. user sign up and UST).

packages/contracts/test/reimburseAttestation.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        const reimbursement = event.args.amount
+
+        const userFinalBalance = await ethers.provider.getBalance(user)
+
+        expect(BigInt(userFinalBalance.toString())).equal(
+            BigInt(userInitialBalance) - BigInt(gasCost) + BigInt(reimbursement)
+        )
+    })
+
+    it('user state transition', async () => {
```
vivianjeng, line 96: It should also check the reimbursement?

packages/contracts/contracts/test/MockUnirep.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {VerifySignature} from '../libraries/VerifySignature.sol';
+import {IUnirep} from '../interfaces/IUnirep.sol';
+
+contract MockUnirep is IUnirep, VerifySignature {
```
vivianjeng, line 7: why use a mock unirep here?

yuriko627: Reference: https://github.com/Unirep/Unirep/pull/470#discussion_r1212496148

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * @dev Add users to whitelist
+     */
+    function addToWhitelist(
+        address[] calldata addresses,
+        address owner
+    ) public virtual {
+        require(
+            msg.sender == owner,
```
vimwitch, line 46: I'm not sure what this check is. It's checking for the sender to be an argument to the function, was this meant to be something else? Like storing `owner` as a state variable?

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * @dev Remove users from whitelist
+     */
+    function removeFromWhitelist(
+        address[] calldata toRemoveAddresses,
+        address owner
+    ) external {
+        require(
+            msg.sender == owner,
```
vimwitch, line 62: Here it also doesn't make sense

packages/contracts/contracts/attester-utils/ReimburseAttestation.sol
vimwitch: I think this was a typo in my writing, the fee would only be reimbursed once

vimwitch (changes requested):

packages/contracts/contracts/attester-utils/ReimburseTransactions.sol
```diff
@@ -0,0 +1,137 @@
...
+        emit Reimbursed(msg.sender, reimbursement, TxType.Signup);
+    }
+
+    function userStateTransition(
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof
+    ) public payable virtual {
+        require(whitelist[msg.sender], 'This user is not in the whitelist');
+
+        uint256 attesterSignalIndex = unirep.numEpochKeyNoncePerEpoch();
```
vivianjeng, line 119: the signal index is not `numEpochKeyNoncePerEpoch`
just hardcode like `uint256 attesterSignalIndex = 2;`

packages/contracts/contracts/test/MockUnirep.sol
```diff
@@ -0,0 +1,56 @@
...
+    // Attester id == address
+    mapping(uint160 => AttesterData) attesters;
+    uint8 public immutable numEpochKeyNoncePerEpoch;
+
+    constructor() {
+        numEpochKeyNoncePerEpoch = 2;
+    }
```
vivianjeng, line 14: it is not used here

vivianjeng (changes requested):

PR #478, opened 2023-06-01
fix(core): prevent running setup twice

vivianjeng:
## Pull request checklist

- [ ] Build (`yarn build`) was run locally in root directory and any changes were pushed
- [ ] Lint (`yarn lint:fix`) was run locally in root directory and any changes were pushed

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/Synchronizer.ts
+++ b/packages/core/src/Synchronizer.ts
@@ -48,6 +48,7 @@ export class Synchronizer extends EventEmitter {
     public blockRate: number = 100000
 
     private setupComplete = false
+    private setupPromise
 
     private lock = new AsyncLock()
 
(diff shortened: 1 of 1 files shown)


packages/core/src/Synchronizer.ts
```diff
@@ -307,7 +307,10 @@ export class Synchronizer extends EventEmitter {
      * block we'll poll many times quickly
      */
     async start() {
-        await this.setup()
+        for (;;) {
```
vimwitch, line 310: I have a different approach [here](https://github.com/Unirep/auth/blob/main/src/Synchronizer.js#L104). I'd rather use a shared promise than poll on this. The same promise is used in many places, [here](https://github.com/Unirep/auth/blob/main/src/Synchronizer.js#L149), [here](https://github.com/Unirep/auth/blob/main/src/Synchronizer.js#L170). So we can check for setup before building trees or doing anything that needs the `settings` object

vimwitch (changes requested):

vimwitch (approved):

PR #485, opened 2023-06-06
Fix attest order

### LINKED ISSUE #450: [Audit warning] Assumed Replacement Data Ordering
The Unirep protocol allows attesters to declare “Replacement Data” with content that will be replaced with each attribution as dictated by the data’s id. In Unirep’s typescript library, though, they assume that the latest attribution is the one with the highest id and therefore the one that should be preserved. While the Unirep developers have switched to a global ID that should match this expectation, it does not consider the case where users of the protocol override the id to provide different behaviors.

> Location where replacement data is set based on attestation order
> 
> 
> ```
>     public getData = async (
>         _toEpoch?: number,
>         _attesterId: bigint | string = this.sync.attesterId
>     ): Promise<bigint[]> => {
>         ...
> 
>         if (orClauses.length === 0) return data
>         const attestations = await this.sync._db.findMany('Attestation', {
>             where: {
>                 OR: orClauses,
>                 attesterId: attesterId,
>             },
>             orderBy: {
>                 index: 'asc',
>             },
>         })
>         for (const a of attestations) {
>             const { fieldIndex } = a
>             if (fieldIndex < this.sync.settings.sumFieldCount) {
>                 data[fieldIndex] = (data[fieldIndex] + BigInt(a.change)) % F
>             } else {
>                 data[fieldIndex] = BigInt(a.change)
>             }
>         }
>         return data
>     }
> ```
> 

## Impact

As the above function is intended for use by users who will be submitting proofs to the ZK circuits, if the correct data is not fetched, users will be unable to submit valid proofs to the ZK circuits.

## Recommendation

While we do note that this is consistent with Unirep’s behavior after the switch to a global ID counter, it deviates from the behavior that Unirep has in their documentation. If this library is intended to be used by those who extend Unirep, the developers should consider slightly modifying this logic to be consistent with the documentation they have available for developers.

vivianjeng:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: fix #450 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/contracts/deploy/deploy.ts
+++ b/packages/contracts/deploy/deploy.ts
@@ -127,18 +127,18 @@ export const deployVerifierHelper = async (
 /**
  * Deploy the unirep contract and verifier contracts with given `deployer` and settings
  * @param deployer A signer who will deploy the contracts
- * @param _settings The settings that the signer can define: `epochLength`, `attestingFee`, `maxUsers`, `maxAttesters`
+ * @param settings The settings that the signer can define: `epochLength`, `attestingFee`, `maxUsers`, `maxAttesters`
  * @returns The Unirep smart contract
  */
 export const deployUnirep = async (
     deployer: ethers.Signer,
-    _settings: CircuitConfig,
+    settings?: CircuitConfig,
     prover?: Prover
 ): Promise<Unirep> => {
     if (!deployer.provider) {
         throw new Error('Deployer must have provider')
     }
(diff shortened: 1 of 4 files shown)


vimwitch (approved):
I'm not sure if this pr fixes #450, we need to ask veridise for clarification

PR #487, opened 2023-06-06
445 synchronization

### LINKED ISSUE #445: Synchronization takes a long time
As time goes on synchronization takes longer. This is because the synchronizer loads sets of `blockRate` blocks in each loop. By default this is 100,000. So after 1 million blocks synchronization will take at least 10 seconds (assuming each load takes 1 second).

On mainnet this would be okay because 1 million blocks = ~139 days. On an L2 like arbitrum blocks are much more frequent so the synchronization takes longer.

We don't want to load too many blocks at once because there might be many events, but we don't want to load too few because then synchronization takes too long. It would be nice if we could request the first `n` events from a range of blocks but that doesn't seem to be possible.

Another approach might be loading the batches in parallel, but this puts more strain on the underlying server.

0xWOLAND:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently, the synchronizer serially loads `blockRate` blocks and processes them in each loop. 

Issue Number:  Closes #445 

## What is the new behavior?

- Blocks are loaded in parallel and separately processed
- Each parallel request uses exponential backoff logic
- `chunks` (Segments of blocks like blocks 1-20) are processed in order

- [ ] Yes
- [x] No

## Other information
Currently, chunks are contiguous sets of 20 blocks

### DIFF
--- a/packages/core/src/Synchronizer.ts
+++ b/packages/core/src/Synchronizer.ts
@@ -23,13 +23,23 @@ type AttesterSetting = {
 
 class BlockQueue {
     private promises: any[] = []
-    public arr: any[] = []
-    public blockEnd: Number = 0
+    private arr: any[] = []
+    public _blockEnd: Number = 0
+
+    get blocks() {
+        return this.arr
+    }
+
+    get blockEnd() {
+        return this._blockEnd
+    }
+
     insert(block: Promise<any>, idx: number) {
         block.then((r) => {
             this.promises.push({ idx, r })
         })
     }
+
     update() {
         this.promises.sort((a, b) => {
             return a.idx - b.idx
@@ -45,6 +55,7 @@ class BlockQueue {
         }
         this.promises = tmp
     }
+
     clear() {
         this.arr = []
         if (this.promises.length > 0) {
@@ -78,7 +89,7 @@ export class Synchronizer extends EventEmitter {
 
     private pollId: string | null = null
     public pollRate: number = 5000
-    public blockRate: number = 100000
+    public blockRate: number = 20
 
     private setupComplete = false
     private setupPromise
@@ -357,7 +368,6 @@ export class Synchronizer extends EventEmitter {
      * block we'll poll many times quickly
      */
     async start() {
-        const N = 20
         await this.setup()
         ;(async () => {
             const pollId = nanoid()
@@ -367,7 +377,7 @@ export class Synchronizer extends EventEmitter {
             for (;;) {
                 // poll repeatedly until we're up to date
                 try {
-                    await this.loadBlocks(N)
+                    await this.loadBlocks(this.blockRate)
                     this.blockQueue.update()
                 } catch (err) {
                     console.error(`--- unable to load blocks`)
@@ -388,7 +398,7 @@ export class Synchronizer extends EventEmitter {
                 if (pollId != this.pollId) break
             }
             for (;;) {
-                await this.loadBlocks(N)
+                await this.loadBlocks(this.blockRate)
                 await new Promise((r) => setTimeout(r, this.pollRate))
                 this.blockQueue.update()
                 if (pollId != this.pollId) break
@@ -432,7 +442,7 @@ export class Synchronizer extends EventEmitter {
         })
         const latestBlock = await this.provider.getBlockNumber()
 
-        const newEvents = this.blockQueue.arr
+        const newEvents = this.blockQueue.blocks
         this.blockQueue.clear()
 
         // filter out the events that have already been seen
@@ -456,12 +466,12 @@ export class Synchronizer extends EventEmitter {
                 OR: this.attestersOrClauses,
             },
             update: {
-                latestCompleteBlock: this.blockQueue.blockEnd,
+                latestCompleteBlock: this.blockQueue._blockEnd,
             },
         })
 
         return {
-            complete: latestBlock === this.blockQueue.blockEnd,
+            complete: latestBlock === this.blockQueue._blockEnd,
         }
     }
 
@@ -479,7 +489,7 @@ export class Synchronizer extends EventEmitter {
         const latestBlock = await this.provider.getBlockNumber()
         const blockStart = latestProcessed + 1
         const count = Math.ceil((latestBlock - blockStart + 1) / n)
-        this.blockQueue.blockEnd = latestBlock
+        this.blockQueue._blockEnd = latestBlock
         if (count <= 0) return
 
         const promises = Array.from(Array(count).keys()).map(async (_, i) => {
--- a/packages/core/src/Synchronizer.ts
+++ b/packages/core/src/Synchronizer.ts
@@ -21,50 +21,6 @@ type AttesterSetting = {
     epochLength: number
 }
 
-class BlockQueue {
-    private promises: any[] = []
-    private arr: any[] = []
-    public _blockEnd: Number = 0
-
-    get blocks() {
-        return this.arr
-    }
-
-    get blockEnd() {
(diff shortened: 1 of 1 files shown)


packages/core/src/Synchronizer.ts
```diff
@@ -21,6 +21,39 @@ type AttesterSetting = {
     epochLength: number
 }
 
+class BlockQueue {
+    private promises: any[] = []
+    public arr: any[] = []
```
vivianjeng, line 26: I think this can be read-only
so we can use a public getter to access the data
https://github.com/Unirep/Unirep/blob/19e489ccaed0392765654ae743f2c2d04f257358/packages/core/src/Synchronizer.ts#L175-L177

packages/core/src/Synchronizer.ts
```diff
@@ -321,6 +357,7 @@ export class Synchronizer extends EventEmitter {
      * block we'll poll many times quickly
      */
     async start() {
+        const N = 20
```
vivianjeng, line 360: it can be public variable like 
https://github.com/Unirep/Unirep/blob/19e489ccaed0392765654ae743f2c2d04f257358/packages/core/src/Synchronizer.ts#L47
so it can be changed in different synchronizer?

0xWOLAND: switched `const N = 20` to use `blockRate` since it wasn't being used elsewhere

packages/core/src/Synchronizer.ts
0xWOLAND: Yep, added

vimwitch (changes requested):
I think this is a good first run. This pr implements parallel block loading, but not concurrent block loading. This diagram should illustrate the difference, each line is a request: 
image

I think parallel loading is okay for now. I'm not sure how much we would gain from more concurrency.

packages/core/src/Synchronizer.ts
```diff
@@ -21,6 +21,50 @@ type AttesterSetting = {
     epochLength: number
 }
 
+class BlockQueue {
```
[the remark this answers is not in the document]
0xWOLAND: merged `BlockQueue` into `Synchronizer`

packages/core/src/Synchronizer.ts
```diff
@@ -407,30 +459,79 @@ export class Synchronizer extends EventEmitter {
...
+        const blockStart = latestProcessed + 1
+        const count = Math.ceil((latestBlock - blockStart + 1) / n)
+        this.blockQueue._blockEnd = latestBlock
+        if (count <= 0) return
+
+        const promises = Array.from(Array(count).keys()).map(async (_, i) => {
+            return this.loadNewEvents(
+                blockStart + n * i,
+                Math.min(blockStart + n * (i + 1) - 1, latestBlock),
+                i
```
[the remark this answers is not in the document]
0xWOLAND: Yep, sorting by `blockNumber` now

packages/core/src/Synchronizer.ts
```diff
@@ -329,6 +376,14 @@ export class Synchronizer extends EventEmitter {
             let backoff = minBackoff
             for (;;) {
                 // poll repeatedly until we're up to date
+                try {
+                    await this.loadBlocks(this.blockRate)
+                    this.blockQueue.update()
```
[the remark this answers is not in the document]
0xWOLAND: Yep, moved to wait for parallel promises to resolve, then update inside of `loadBlocks`

packages/core/src/Synchronizer.ts
```diff
@@ -316,6 +321,41 @@ export class Synchronizer extends EventEmitter {
...
 
+    get blocks() {
+        return this._blocks
+    }
+
+    get blockEnd() {
+        return this._blockEnd
+    }
+
+    insert(block: Promise<any>) {
```
vimwitch, line 332: this function is only being used in 1 place, i think the logic would fit better in the `loadNewEvents` function

packages/core/src/Synchronizer.ts
```diff
@@ -316,6 +321,41 @@ export class Synchronizer extends EventEmitter {
...
+        block.then((r) => {
+            this.promises.push(r)
+        })
+    }
+
+    update() {
```
vimwitch, line 338: same here, the logic can be put in the `loadBlocks` function

packages/core/src/Synchronizer.ts
```diff
@@ -316,6 +321,41 @@ export class Synchronizer extends EventEmitter {
...
+                continue
+            }
+            for (const block of chunk) {
+                this._blocks.splice(0, 0, block)
+            }
+        }
+        this.promises = tmp
+    }
+
+    clear() {
```
vimwitch, line 355: same here

packages/core/src/Synchronizer.ts
```diff
@@ -343,6 +390,8 @@ export class Synchronizer extends EventEmitter {
                 if (pollId != this.pollId) break
             }
             for (;;) {
+                console.log(`${pollId} - ${this.pollId}`)
```
vimwitch, line 393: stray console

packages/core/src/Synchronizer.ts
```diff
@@ -316,6 +321,41 @@ export class Synchronizer extends EventEmitter {
...
+        block.then((r) => {
+            this.promises.push(r)
+        })
+    }
+
+    update() {
+        this.promises.sort((a, b) => {
+            return a.blockNumber - b.blockNumber
+        })
+        console.log(this.promises)
```
vimwitch, line 342: stray console

packages/core/src/Synchronizer.ts
```diff
@@ -316,6 +321,41 @@ export class Synchronizer extends EventEmitter {
         })
     }
 
+    get blocks() {
```
vimwitch, line 324: I don't think we need getters here, these variables should not be accessed outside of this class

0xWOLAND: removed

packages/core/src/Synchronizer.ts
0xWOLAND: removed

packages/core/src/Synchronizer.ts
0xWOLAND: removed

packages/core/src/Synchronizer.ts
0xWOLAND: moved function into `loadBlocks`

packages/core/src/Synchronizer.ts
0xWOLAND: moved logic to `loadBlocks`

packages/core/src/Synchronizer.ts
0xWOLAND: move logic to `loadNewEvents`

packages/core/src/Synchronizer.ts
```diff
@@ -45,13 +45,17 @@ export class Synchronizer extends EventEmitter {
 
     private pollId: string | null = null
     public pollRate: number = 5000
-    public blockRate: number = 100000
+    public blockRate: number = 20
```
vivianjeng, line 48: this value can be set higher e.g. `100000`
or the memory will be run out easily
```shell
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

0xWOLAND: fixed

vivianjeng (approved):

PR #496, opened 2023-06-21
Update benchmark scripts

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Issue Number: N/A

## What is the new behavior?

- use `yarn circuits time` to check benchmark script
- The parameters could be changed in `packages/circuits/scripts/time.mjs`
-

- [ ] Yes
- [x] No


.circleci/config.yml
```diff
@@ -102,6 +102,9 @@ jobs:
                   command: |
                       diff $ABI_ORIGIN $ABI_COPY
                       rm $ABI_COPY
+            - run:
```
vimwitch, line 105: I don't think we should run this in CI on every build. It's adding ~2.5 minutes to the ci time

vimwitch:
I still think we shouldn't execute the benchmark in CI. Anytime we make changes to the circuits I think we could run the benchmark locally and post the results in a comment in the PR.

Another option would be benchmarking the pre-built circuits from the build step. In the script we're benchmarking the max tree size config - I'm not sure how useful that is

vivianjeng:
I think because we don't have the CI for benchmark so we didn't maintain this file for a long time 😅
Or we could print out the constraints when building circuits and remove the `time.ts` file
(and the proving time is also printed out in circuits tests)

PR #526, opened 2023-08-04
add Matomo for web tracking

ChialiT:
Adding web tracking to our site, using the method from : https://github.com/karser/docusaurus-plugin-matomo


ChialiT:
i think i didnt complete the lint.

CJ-Rose:
@ChialiT we are using Yarn package manager so if you try to install something with npm it messes up the build pretty bad. i tried to remove it and install with yarn but i think it was too confused to repair so i made a new PR with your changes #527. 
(the command is always `yarn add <package>` instead of `npm install <package>`)
not sure if it's doing what it's supposed to 😅 but if it is working, check out how the additions to `docusaurus.config` are inserted into the existing structure instead of sticking on to the end.

PR #536, opened 2023-08-29
Add UniRepVoting as an example

joyqvq:
Add an example of voting with Unirep voting, the user can cast vote out of numOptions via reputation proof. 

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): Example

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/packages/contracts/contracts/examples/UnirepVoting.sol
+++ b/packages/contracts/contracts/examples/UnirepVoting.sol
@@ -1,0 +1,167 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+import {Unirep} from '.././Unirep.sol';
+
+// Uncomment this line to use console.log
+// import 'hardhat/console.sol';
+import {ReputationVerifierHelper} from '../verifierHelpers/ReputationVerifierHelper.sol';
+import {EpochKeyVerifierHelper} from '../verifierHelpers/EpochKeyVerifierHelper.sol';
+import {VotingPrizeNFT} from './VotingPrizeNFT.sol';
+
+enum Option {
+    UP,
+    DOWN
+}
+
+contract UnirepVoting {
+    Unirep public unirep;
+    ReputationVerifierHelper public helper;
+    EpochKeyVerifierHelper public epochKeyHelper;
+
+    int[] public scores;
+    uint public immutable numTeams;
+    int public winnerScore;
+    uint[][] public projectData;
+    bool foundWinner = false;
+    VotingPrizeNFT public nft;
+
+    mapping(uint256 => uint256[]) public participants;
+    mapping(uint256 => uint256) public counts;
+    mapping(uint256 => uint256) public voted;
+    mapping(uint256 => bool) claimed;
+
+    constructor(
+        Unirep _unirep,
+        ReputationVerifierHelper _helper,
+        EpochKeyVerifierHelper _epochKeyHelper,
+        VotingPrizeNFT _nft,
+        uint8 _numTeams,
+        uint48 _epochLength
+    ) {
+        // set unirep address
+        unirep = _unirep;
+
+        // set verifier address
+        helper = _helper;
+        epochKeyHelper = _epochKeyHelper;
+
+        nft = _nft;
+
+        unirep.attesterSignUp(_epochLength);
+
+        // how many numTeams
+        numTeams = _numTeams;
+        scores = new int[](numTeams);
+        projectData = new uint[][](numTeams);
+        for (uint i; i < numTeams; i++) {
+            projectData[i] = new uint256[](4);
+        }
+    }
+
+    // sign up users in this app
+    function userSignUp(
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof
+    ) public {
+        unirep.userSignUp(publicSignals, proof);
+    }
+
+    function joinProject(
+        uint256 projectID,
+        uint256[] memory publicSignals,
+        uint256[8] memory proof
+    ) public {
+        require(projectID < numTeams, 'Voteathon: invalid project iD');
+        EpochKeyVerifierHelper.EpochKeySignals memory signals = epochKeyHelper
+            .verifyAndCheck(publicSignals, proof);
+
+        require(signals.revealNonce == true, 'Voteathon: should reveal nonce');
+        require(signals.nonce == 0, 'Voteathon: invalid nonce');
+        require(
+            counts[projectID] < 10,
+            'Voteathon: maximum participants in a project'
+        );
+        participants[projectID].push(signals.epochKey);
+        // give user data if there is attestation before
+        uint48 epoch = unirep.attesterCurrentEpoch(uint160(address(this)));
+        require(epoch == 0, 'Voteathon: not join epoch');
+        uint256[] memory data = projectData[projectID];
+        for (uint256 i = 0; i < data.length; i++) {
+            unirep.attest(signals.epochKey, epoch, i, data[i]);
+        }
+        counts[projectID] += 1;
+    }
+
+    function vote(
+        uint256 projectID,
+        Option option,
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof
+    ) public {
+        require(projectID < numTeams, 'projectID out of range');
+
+        EpochKeyVerifierHelper.EpochKeySignals memory signals = epochKeyHelper
+            .verifyAndCheck(publicSignals, proof);
+
+        require(signals.revealNonce == true, 'reveal nonce wrong');
+        require(signals.nonce == 1, 'nonce wrong');
+        require(signals.epoch == 0, 'invalid epoch to vote');
+
+        projectData[projectID][uint(option)] += 1;
+
+        voted[signals.epochKey] += 1;
+        if (option == Option.UP) scores[projectID] += 1;
+        else if (option == Option.DOWN) scores[projectID] -= 1;
+
+        uint[] memory members = participants[projectID];
+        uint48 epoch = unirep.attesterCurrentEpoch(uint160(address(this)));
+        require(epoch == 0, 'not voting epoch');
+        for (uint256 i = 0; i < members.length; i++) {
+            unirep.attest(members[i], epoch, uint(option), 1);
+        }
+
+        uint160 attesterId = uint160(address(this));
+        unirep.attesterStateTreeRootExists(
+            attesterId,
+            epoch,
+            signals.stateTreeRoot
+        );
+    }
+
+    function claimPrize(
+        address receiver,
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof
+    ) public {
+        uint160 attesterId = uint160(address(this));
+        require(unirep.attesterCurrentEpoch(attesterId) > 0);
+        ReputationVerifierHelper.ReputationSignals memory signals = helper
+            .verifyAndCheck(publicSignals, proof);
+        uint48 epoch = 0; // the voting epoch
+        uint256 stateTreeRoot = signals.stateTreeRoot;
+        unirep.attesterStateTreeRootExists(attesterId, epoch, stateTreeRoot);
+        require(signals.revealNonce == true, 'reveal nonce wrong');
+        require(signals.nonce == 1, 'nonce wrong');
+        require(signals.epoch > 0, 'invalid epoch to claim prize');
+
+        require(!claimed[signals.epochKey], 'Already claimed');
+        if (!foundWinner) {
+            _findWinner();
+        }
+        require(int(signals.minRep) >= winnerScore, 'Insufficient score');
+        nft.awardItem(receiver);
+        claimed[signals.epochKey] = true;
+    }
+
+    function _findWinner() internal {
+        int highest = 0;
+        for (uint256 i = 0; i < numTeams; i++) {
+            if (scores[i] > highest) {
+                highest = scores[i];
+            }
+        }
+        winnerScore = highest;
+        foundWinner = true;
+    }
+}
--- a/packages/contracts/contracts/examples/UnirepVoting.sol
+++ b/packages/contracts/contracts/examples/UnirepVoting.sol
@@ -142,9 +142,13 @@ contract UnirepVoting {
         require(signals.epoch > 0, 'invalid epoch to claim prize');
         require(signals.revealNonce == true, 'reveal nonce wrong');
         require(signals.nonce == 1, 'nonce wrong');
-        
+
         uint256 stateTreeRoot = signals.stateTreeRoot;
-        unirep.attesterStateTreeRootExists(attesterId, signals.epoch, stateTreeRoot);
+        unirep.attesterStateTreeRootExists(
+            attesterId,
+            signals.epoch,
+            stateTreeRoot
+        );
 
         require(!claimed[signals.epochKey], 'Already claimed');
         if (!foundWinner) {
--- a/packages/contracts/contracts/examples/UnirepVoting.sol
+++ b/packages/contracts/contracts/examples/UnirepVoting.sol
@@ -16,7 +16,7 @@ enum Option {
 
 contract UnirepVoting {
     Unirep public unirep;
-    ReputationVerifierHelper public helper;
+    ReputationVerifierHelper public repHelper;
     EpochKeyVerifierHelper public epochKeyHelper;
 
     int[] public scores;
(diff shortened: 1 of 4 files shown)


packages/contracts/test/unirepVoting.test.ts
```diff
@@ -0,0 +1,111 @@
...
+                'dataProof',
+                circuitInputs
+            )
+            const dataProof = new DataProof(
+                p.publicSignals,
+                p.proof,
+                defaultProver
+            )
+            console.log('option', option)
+            console.log('num option', await voting.numOptions())
```
joyqvq, line 103: i initialized the numOptions as 10 at the beginning, but somehow here it prints 0 🤔

vimwitch: I think you are assigning `numOptions` incorrectly in the constructor [here](https://github.com/Unirep/Unirep/pull/536/files#diff-93396ed1d44806f3139099c801bbf68cf0ba914268c9d63234f4dee3a2af93e0R37)

joyqvq: ahh yes , thanks!

vivianjeng:
I think you can just add the example file in 
`packages/contracts/contracts/examples`
so it can be easily found by developers

joyqvq:
> I think you can just add the example file in `packages/contracts/contracts/examples` so it can be easily found by developers

its added here: https://github.com/Unirep/Unirep/blob/ad9b58e780af28dc18c0aaa9657d230c4d21f04a/packages/contracts/contracts/examples/UnirepVoting.sol

you mean other files re circuits too?

packages/circuits/circuits/dataProof.circom
```diff
@@ -0,0 +1,65 @@
...
+    signal input attester_id;
+    signal input data[FIELD_COUNT];
+    signal output state_tree_root;
+    signal input epoch;
+
+    // Prove values
+    signal input value[SUM_FIELD_COUNT];
+
+    // Values that indicate if the user has claimed before.
+    signal output nullifier;
```
joyqvq, line 26: i still want to use the output to indicate whether someone has voted for the option. 

i copied this circuit from voteathon, i think the dataProof can be reused without the "claim nft" part, this circuit is to check: 
1) the voter had NOT voted for the option before
2) the voter's vote is within the range of options

joyqvq: error and printed public signals: 
```
  publicSignals: [
    4928033182956221957602607196326425475980978144784113335039380808473053488197n,
    6473017752668400910032266656873353403831686455175612966358707049578912557254n,
    0n,
    0n,
    0n,
    0n
  ],
  prover: {
    genProofAndPublicSignals: [AsyncFunction: genProofAndPublicSignals],
    verifyProof: [AsyncFunction: verifyProof],
    getVKey: [AsyncFunction: getVKey]
  },
  idx: { stateTreeRoot: 0, value: [ 1, 5 ] },
  stateTreeRoot: 4928033182956221957602607196326425475980978144784113335039380808473053488197n,
  value: [
    6473017752668400910032266656873353403831686455175612966358707049578912557254n,
    0n,
    0n,
    0n
  ],
  circuit: 'dataProof'
}
vote 4928033182956221957602607196326425475980978144784113335039380808473053488197
    1) only signed up voter can vote exactly once


  2 passing (13s)
  1 failing

  1) Voting
       only signed up voter can vote exactly once:
     Error: Transaction reverted: function returned an unexpected amount of data
```

packages/contracts/contracts/examples/UnirepVoting.sol
```diff
@@ -0,0 +1,76 @@
...
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof
+    ) public {
+        unirep.userSignUp(publicSignals, proof);
+    }
+
+    function vote(
+        uint256 option,
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof
```
vivianjeng, line 50: I think it can be epoch key proof here?
just to make sure the user has signed up in this app
and we can reveal the epoch key nonce (e.g. `nonce = 0`) to make sure the user won't vote many times

packages/contracts/contracts/examples/UnirepVoting.sol
```diff
@@ -0,0 +1,76 @@
...
+        require(option < numOptions, 'voted option out of range');
+
+        ReputationVerifierHelper.ReputationSignals
+            memory signals = decodeAndVerify(publicSignals, proof);
+
+        require(voted[signals.epochKey] < 1, 'Already voted');
+
+        if (signals.maxRep > 0) {
+            voted[signals.epochKey] += 1;
+        }
```
vivianjeng, line 61: I think we don't need the check here.

packages/contracts/contracts/examples/UnirepVoting.sol
```diff
@@ -0,0 +1,76 @@
...
+        scores[option] += 1;
+
+        uint48 epoch = unirep.attesterCurrentEpoch(uint160(address(this)));
```
vivianjeng, line 64: and we also need to check the `stateTreeRoot` in the proof (either epoch key proof or reputation proof)
to make sure the user has signed up in this system.

packages/contracts/contracts/examples/UnirepVoting.sol
```diff
@@ -0,0 +1,76 @@
...
+        unirep.attesterSignUp(_epochLength);
+    }
+
+    // sign up users in this app
+    function userSignUp(
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof
+    ) public {
+        unirep.userSignUp(publicSignals, proof);
+    }
```
vivianjeng, line 45: and I think we should let the user generate epoch key proof when they sign up?
so users can receive votes (pretty much like Voteathon)
and they can join projects

packages/contracts/contracts/examples/UnirepVoting.sol
```diff
@@ -0,0 +1,76 @@
...
+
+    // decode and verify the proofs
+    // fails or returns proof signals
+    function decodeAndVerify(
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof
+    ) public view returns (ReputationVerifierHelper.ReputationSignals memory) {
+        return helper.verifyAndCheck(publicSignals, proof);
+    }
```
vivianjeng, line 75: It can just call the `helper.verifyAndCheck(publicSignals, proof);`
we don't need the extra function?

packages/contracts/package.json
```diff
@@ -21,7 +21,7 @@
         "build": "ts-node ./scripts/buildVerifiers.ts && yarn compile && yarn abi && hardhat typechain && tsc",
         "abi": "node ./scripts/abi",
         "compile": "hardhat compile",
-        "test": "hardhat test ./test/*.test.ts",
+        "test": "hardhat test ./test/unirepVoting.test.ts",
```
vivianjeng, line 24: please revert the change of this line 🙏🏻

vivianjeng (changes requested):
I think we can still have reputation proof to claim prize
just use [`userState.genProveReputationProof({minRep: winnerScore})`](https://developer.unirep.io/docs/core-api/user-state#genprovereputationproof)
and maybe need to reveal nonce to make sure the user doesn't claim prize more than once

packages/contracts/contracts/examples/UnirepVoting.sol
joyqvq: hm should user submit 2 proofs? one reputation and one epoch key proof?

vivianjeng: ah I think it could be `signup` and `joinProject`
and signup can be for hackers and mentors and others
and joinProject is only for hackers
and `signup` can take the signup proof
and `joinProject` can take epoch key proof

packages/contracts/contracts/examples/UnirepVoting.sol
joyqvq: yeah this is the epoch key proof to sign up with. instead of users can join projects, i simplified by just having options to vote for (numOptions)

packages/contracts/contracts/examples/UnirepVoting.sol
joyqvq: ok ill try implement this - i had a different version in my implementation!

joyqvq: without dataProof, how do you check if a user is part of a project or not?

packages/contracts/package.json
joyqvq: whats the best way to just run one test? i tried `hardhat test ./test/unirep.test.ts` but got command not found

packages/contracts/contracts/examples/UnirepVoting.sol
joyqvq: i revised it to be: 

joinProject with epoch key proof
vote takes epoch key proof
claim prize takes reputation proof (im not sure which field is useful here, is it minRep and maxRep? what are they?)

packages/contracts/contracts/examples/UnirepVoting.sol
joyqvq: added
```
        uint160 attesterId = uint160(address(this));
        unirep.attesterStateTreeRootExists(attesterId, epoch, signals.stateTreeRoot);
```

packages/contracts/contracts/examples/UnirepVoting.sol
vivianjeng: > without dataProof, how do you check if a user is part of a project or not?

just use reputation proof
since the user will receive data from every vote
(so we can protect user's privacy, he doesn't have to reveal the project he joined)

> claim prize takes reputation proof (im not sure which field is useful here, is it minRep and maxRep? what are they?)
Yes the `minRep` field
It checks if `data[0]-data[1] >= minRep` or not

https://github.com/Unirep/Unirep/blob/67b2b5989b69c52568dda0b7e478e92ffadb83d8/packages/circuits/circuits/proveReputation.circom#L102

packages/contracts/package.json
vivianjeng: you can either change the `test` script in your local to be like this (but not commit this change)
or you can run `hardhat test $FILE_PATH` in the root `packages/contracts` or `packages/core`
since in root directory there is no `hardhat` installed

packages/contracts/contracts/examples/UnirepVoting.sol
```diff
@@ -0,0 +1,148 @@
...
+    
+    function claimPrize(
+        address receiver,
+        uint256[] calldata publicSignals,
+        uint256[8] calldata proof
+    ) public {
+        uint160 attesterId = uint160(address(this));
+        require(unirep.attesterCurrentEpoch(attesterId) > 0);
+        ReputationVerifierHelper.ReputationSignals memory signals = helper.verifyAndCheck(publicSignals, proof);
+        uint48 epoch = 0; // the voting epoch
```
vivianjeng, line 124: It should be the next epoch that the user transitions to. (`epoch > 0`)
There should be no data before user transition (`epoch=0`)
and we want to verify the latest proof (`epoch>0`)

packages/contracts/contracts/examples/UnirepVoting.sol
```diff
@@ -0,0 +1,148 @@
...
+        uint256 stateTreeRoot = signals.stateTreeRoot;
+        unirep.attesterStateTreeRootExists(attesterId, epoch, stateTreeRoot);
+        require(!claimed[publicSignals[1]], 'Already claimed');
```
vivianjeng, line 127: we can also verify the `revealNonce` and `nonce` in this function.

```solidity
        require(signals.revealNonce == true, "reveal nonce wrong");
        require(signals.nonce == 1, "nonce wrong");
```
to make sure the user doesn't use two epoch keys to receive prize
and there should be a `signals.epochKey` in [`ReputationSignals`](https://developer.unirep.io/docs/contracts-api/verifiers/reputation-verifier-helper) you can use instead of using `publicSignals[1]`

packages/contracts/contracts/examples/UnirepVoting.sol
```diff
@@ -0,0 +1,148 @@
...
+        int score = int(publicSignals[2]) + int(publicSignals[3]);
```
vivianjeng, line 128: maybe this line is not needed in this function.
instead we can check `signals.minRep >= winnerScore`

packages/contracts/contracts/examples/UnirepVoting.sol
joyqvq: this epoch 0 is used to check: unirep.attesterStateTreeRootExists(attesterId, epoch, stateTreeRoot);

this is what the epoch that vote function is called: 
```
        require(epoch == 0, 'not voting epoch');
        for (uint256 i = 0; i < members.length; i++) {
            unirep.attest(members[i], epoch, uint(option), 1);
        }
```

vivianjeng: ah then it should still check `epoch>0`
```solidity
require(epoch > 0, 'should be after voting epoch');
uint256 stateTreeRoot = signals.stateTreeRoot;
unirep.attesterStateTreeRootExists(attesterId, epoch, stateTreeRoot);
```

joyqvq: i think this is what you mean?

1) in `claim_prize`: check the epoch in the reputation proof for should > 0 
2) in `vote`: the epoch in the epoch key proof epoch should  = 0

revised to include them

packages/contracts/contracts/examples/UnirepVoting.sol
```diff
@@ -0,0 +1,167 @@
...
+        uint160 attesterId = uint160(address(this));
+        require(unirep.attesterCurrentEpoch(attesterId) > 0);
+        ReputationVerifierHelper.ReputationSignals memory signals = helper
+            .verifyAndCheck(publicSignals, proof);
+        uint48 epoch = 0; // the voting epoch
+        uint256 stateTreeRoot = signals.stateTreeRoot;
+        unirep.attesterStateTreeRootExists(attesterId, epoch, stateTreeRoot);
+        require(signals.revealNonce == true, 'reveal nonce wrong');
+        require(signals.nonce == 1, 'nonce wrong');
+        require(signals.epoch > 0, 'invalid epoch to claim prize');
```
joyqvq, line 146: check epoch for claim prize

packages/contracts/test/unirepVoting.test.ts
```diff
@@ -0,0 +1,209 @@
...
+        attesterId,
+        unirepAddress,
+        provider: ethers.provider,
+        id,
+        prover: defaultProver,
+    })
+    await userState.sync.start()
+    await userState.waitForSync()
+    return userState
+}
```
vivianjeng, line 33: I think it should not use `@unirep/core` in `@unirep/contracts`
because `@unirep/core` depends on `@unirep/contracts`

so we should bootstrap signup circuits, epochkey circuits, reputation circuits like other tests
another solution is that we put this test in `@unirep/core` and keep the vote contract here
so we can use `UserState` and its methods

packages/contracts/contracts/examples/UnirepVoting.sol
```diff
@@ -0,0 +1,172 @@
...
+        VotingPrizeNFT _nft,
+        uint8 _numTeams,
+        uint48 _epochLength
+    ) {
+        // set unirep address
+        unirep = _unirep;
+
+        // set verifier address
+        helper = _helper;
+        epochKeyHelper = _epochKeyHelper;
```
vivianjeng, line 47: if there are two helpers, I think reputation helper should be called `_repHelper`

packages/contracts/contracts/examples/UnirepVoting.sol
```diff
@@ -0,0 +1,172 @@
...
+        nft = _nft;
+
+        unirep.attesterSignUp(_epochLength);
+
+        // how many numTeams
+        numTeams = _numTeams;
+        scores = new int[](numTeams);
+        projectData = new uint[][](numTeams);
+        for (uint i; i < numTeams; i++) {
+            projectData[i] = new uint256[](4);
```
vivianjeng, line 58: If there are only two options, `projectData` can be `length=2`?

vivianjeng (commented):
There are some small comments 
But I think it is pretty ready to be merged 🙏🏻

packages/contracts/test/unirepVoting.test.ts
joyqvq: it seems like other tests involves user state lives in unirep/core makes sense that it belongs there. however do you know how getContractFactory works? after moving the contract cannot seem to be found by the test

joyqvq: moved it to package core!

vivianjeng (approved):
LGTM 
thank you

PR #547, opened 2023-09-20
Add Chain ID in proofs

### LINKED ISSUE #139: Include the chain id in the epoch key
If we include the chain id of the unirep instance in the zk proof we can make crosschain proofs.

e.g. I have 20 posRep from attester `0xabababababababababababababababababababab` on the optimism testnet (chain id: 0x1a4).

On mainnet there is a contract issuing poaps for people that have at least 5 rep for the attester on the optimism testnet.

A reputation proof proves the chainId, then during onchain verification the chain id is compared to the value in the evm.

epoch keys could be calculated as `H(idNullifier, chainId, attesterId, epoch, nonce) % tree_leaf_count`

**This would help prevent crosschain replay attacks.**

### LINKED ISSUE #531: Better graffiti handler
Now the circuit takes `repl_data + repl_nonce` as input
and the [`genProveReputationProof`](https://github.com/Unirep/Unirep/blob/dbb235c193670bdd416e9a84470980a3927df919/packages/core/src/UserState.ts#L541) takes only `repl_data` as inputs.
but the [`ReputationProof`](https://github.com/Unirep/Unirep/blob/dbb235c193670bdd416e9a84470980a3927df919/packages/circuits/src/ReputationProof.ts#L73C1-L74C1) outputs `repl_data + repl_nonce`
e.g. 
If I want to prove graffiti = 3 (`repl_data`)
It will output the `proof.graffiti = 300...0000001` where `repl_data + repl_nonce`

See more information in [protocol/data](https://developer.unirep.io/docs/protocol/data)

### LINKED ISSUE #541: Update semaphore v4
Simplified identity: The two old identity secrets (`trapdoor` and `nullifier`) have been replaced by 1 secret only (`secret`). So `identity commitment = hash(identity secret)`, and `nullifier hash = hash(scope, identity secret)`.

We can remove all identity `trapdoor` and `nullifier` to use only `secret`

vivianjeng:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: close #139 close #541 close #531 

## What is the new behavior?

- [**Semaphore v4**](https://pse-team.notion.site/Semaphore-V4-Summary-478b2813bf6d4fdc917661d95885a968)
    - External Nullifier → Scope
    - `nullifier hash = hash(scope, identity secret)`
- Rename circuits
    - `PreventDoubleAction` -> `ScopeNullifier` (to match semaphore v4)
    - `ProveReputation` -> `Reputation` (remove redundant word)
- chain id
    - 36 bits, the maximum safe bit number in one control 
      (epochKey control: 160(attesterId)+48(epoch)+8(nonce)+1(revealNonce)+36(chain)=252bits) 
    - Include chain id in `signup`, `epochKey`, `epochKeyLite`, `userStateTransition`, `reputation`, `scopeNullifier` proofs 
    - `genEpochKey`, `genIdentityHash`, `genStateTreeLeaf` add chain ID
    - `Unirep.sol` checks chain ID in signup proofs
    - add helper on-chain `unirepChainId()`
    - add error `ChainIdNotMatch`
    - update `decodeSignupControl` and `decodeSignupSignals` to decode chain ID
    - update `UserState` to take chain ID to make proofs
- graffiti
    - previously we take both `repl_data` and `repl_nonce` in a graffiti
       but we only need to check `repl_data` to make it easier for users/developers
    - in circuits, we check if `IsEqual()([graffiti, repl_data])`, 
       where `graffiti` and `repl_data` are `253-REPL_NONCE_BITS`
- shift bits
    - previously we use a single signal (variable) to check a control
    - The proposed method is shift and add (both control and accumulated bits)
- test
    - Update tests and let it take the global `circuitInputs` instead of generating `circuitInputs` in each unit, overwrite global `circuitInputs` to take each condition

- [x] Yes
- [ ] No


packages/circuits/README.md
```diff
@@ -106,7 +106,7 @@ import { Circuit } from '@unirep/circuits'
 // See ./test/verifyEpochKey.test.ts for generating circuit inputs
 const circuitInputs = {
     state_tree_elements: ...,
-    state_tree_indexes: ...,
+    state_tree_indeces: ...,
```
vimwitch, line 109: I think this should be either `indexes` or `indices`

packages/circuits/circuits/hasher.circom
```diff
@@ -3,13 +3,38 @@ include "./circomlib/circuits/poseidon.circom";
...
+
+  /**
+  * control structure
+  * 160 bits attester id
+  * 48 bits epoch
+  * 8 bits nonce
+  * 36 bit chain id
+  **/
+  var acc_bits = 0;
+  var acc_data = attester_id;
```
vimwitch, line 26: Perhaps we should add a note that the range of these values should be constrained elsewhere.

packages/circuits/circuits/identity.circom
```diff
@@ -3,10 +3,10 @@ pragma circom 2.1.0;
 include "./circomlib/circuits/poseidon.circom";
 
 template IdentityCommitment() {
-  signal input secret;
```
vimwitch, line 6: This was named secret to avoid duplicate text in circuits.

e.g.

```
template identity_commitment;
identity_commitment.secret <== value;
// instead of
identity_commitment.identity_secret <== value;
```

packages/circuits/src/BaseProof.ts
```diff
@@ -13,8 +12,8 @@ export class BaseProof {
     readonly _snarkProof: SnarkProof
     protected circuit?: Circuit
 
-    readonly publicSignals: bigint[]
-    public proof: bigint[]
+    readonly publicSignals: string[]
+    public proof: string[]
```
vimwitch, line 16: Why are we changing this type from `bigint[]` to `string[]`?

packages/circuits/src/utils.ts
```diff
@@ -0,0 +1,256 @@
...
+    data: string,
+    shiftBits: bigint,
+    variableBits: bigint
+): bigint => {
+    const _data = BigInt(data)
+    return (_data >> shiftBits) & ((BigInt(1) << variableBits) - BigInt(1))
+}
+
+export const decodeEpochKeyControl = (
+    control: string,
```
vimwitch, line 58: Could we take a `bigint` here and in the other functions? Changing the types on the `Proof` classes public properties is a breaking change.

packages/contracts/contracts/verifierHelpers/BaseVerifierHelper.sol
```diff
@@ -8,64 +8,93 @@ contract BaseVerifierHelper {
 
     uint256 public constant SNARK_SCALAR_FIELD =
         21888242871839275222246405745257275088548364400416034343698204186575808495617;
+    uint256 public immutable chainid;
 
     struct EpochKeySignals {
-        bool revealNonce;
```
vimwitch, line 13: Try to keep the smaller types near each other in the `struct` definition. If the types can fit into a single 32 byte area the compiler will do so automatically and use less memory.

e.g.

```
bool a;
uint48 b;
uint160 c;
uint256 d;
```

instead of 

```
bool a;
uint48 b;
uint256 d;
uint160 c;
```

packages/contracts/contracts/verifierHelpers/BaseVerifierHelper.sol
```diff
@@ -8,64 +8,93 @@ contract BaseVerifierHelper {
...
+        accBits += attesterIdBits;
+
+        revealNonce = bool(shift(control, accBits, revealNonceBit) != 0);
+        accBits += revealNonceBit;
+
+        chainId = uint48(shift(control, accBits, chainIdBits));
+        return (nonce, epoch, attesterId, revealNonce, chainId);
+    }
+
+    function shift(
```
vimwitch, line 84: Maybe call this `shiftParse` or `shiftAndParse`?

vimwitch (changes requested):
This is a good pr, but introduces lots of non-backward compatible changes. I think we should release this change under version `2.1.0` e.g. a new minor version.

packages/circuits/README.md
vivianjeng: I think it should match the epoch tree and history tree
https://github.com/Unirep/Unirep/blob/508ebd6c8f3366f923a1657c0800dad2a5a7eb45/packages/circuits/circuits/userStateTransition.circom#L53
https://github.com/Unirep/Unirep/blob/508ebd6c8f3366f923a1657c0800dad2a5a7eb45/packages/circuits/circuits/userStateTransition.circom#L36

packages/circuits/circuits/identity.circom
vivianjeng: I am trying to align the names in all circuits
so it won't confuse developers when they use different circuits with different input signals

packages/circuits/src/BaseProof.ts
vivianjeng: now when we developing attester apps (frontend)
it requires to change types to send proofs to backend
I think `string` is more convenient and it is originally `string` when proof is generated

https://github.com/Unirep/create-unirep-app/blob/833685b2b8b15a63ddbe114b79f2a333e3f0bf43/packages/frontend/src/contexts/User.ts#L160-L165

packages/contracts/contracts/verifierHelpers/BaseVerifierHelper.sol
vivianjeng: [3639c48](https://github.com/Unirep/Unirep/pull/547/commits/3639c48267119900a32eff5b2b2b2c26d64d3241)

packages/circuits/circuits/hasher.circom
vivianjeng: or we can move the constraints to hasher templates?

packages/contracts/contracts/verifierHelpers/BaseVerifierHelper.sol
vivianjeng: I tried several combinations
the current one is the best ( but not sure what it is the optimized)
| | before |  [bc03229](https://github.com/Unirep/Unirep/pull/547/commits/bc03229685101249640c3442bb703cc989e971ec)|
| :--:| :--:| :--:|
| Unirep.sol | 3612695 | 3599327 |
| EpochKeyVerifierHelper.sol | 616423 | 615775 |
| EpochKeyLiteVerifierHelper.sol | 608769 | 607695 |
| ReputationVerifierHelper.sol | 780606 | 772772 |

vivianjeng:
@vimwitch so let's publish the current main branch as `2.0.0` and use this `2.1.0` to perform trusted setup?

packages/circuits/circuits/hasher.circom
vivianjeng: Add notes here, [74ffc42](https://github.com/Unirep/Unirep/pull/547/commits/74ffc4293cca0ae8aad032b819c5855e0cb13087)
I think constraints could be added before using the template (to save constraints)

(the documentation hasn't updated for the hasher templates,
if the documentation is updated, we can also add notes there)

packages/circuits/src/BaseProof.ts
vimwitch: I tend to use `bigint` values everywhere when i work with this library. I think `string` values are ambiguous, they can be e.g. decimal or hex or binary. It's kinda okay because the `string` must be a valid input to `BigInt(str)`, but this is only checked when/if `bigint` conversion happens.

I also think there may be a cost to parsing the `string` into a `bigint`, e.g. serialization.

vimwitch:
I don't think we should include this in `2.0.0`. This pr has quite a lot of changes and we haven't tested it with external applications. The changes also haven't been audited afaik, I think these are a lot of changes to ask Veridise to review without more payment.

vivianjeng:
> I don't think we should include this in `2.0.0`. This pr has quite a lot of changes and we haven't tested it with external applications. The changes also haven't been audited afaik, I think these are a lot of changes to ask Veridise to review without more payment.

@vimwitch I think many of them are minor changes in circuits (except for chain id itself)
and we just need to fix circuits before trusted setup
(so e.g. the circuit interface, contract... we can change later)

Let's ask Veridise for help and wait their review?

packages/circuits/src/utils.ts
vivianjeng: [1658f01](https://github.com/Unirep/Unirep/pull/547/commits/1658f01f89d47cf25d29314f24ce83aa24a5fddf)

packages/circuits/src/BaseProof.ts
vivianjeng: [1658f01](https://github.com/Unirep/Unirep/pull/547/commits/1658f01f89d47cf25d29314f24ce83aa24a5fddf)

packages/contracts/contracts/verifierHelpers/BaseVerifierHelper.sol
```diff
@@ -8,64 +8,95 @@ contract BaseVerifierHelper {
...
+        revealNonce = bool(
+            shiftAndParse(control, accBits, revealNonceBit) != 0
         );
-        epoch = uint48((control >> nonceBits) & ((1 << epochBits) - 1));
-        nonce = uint8(control & ((1 << nonceBits) - 1));
-        return (revealNonce, attesterId, epoch, nonce);
+        accBits += revealNonceBit;
+
+        chainId = uint48(shiftAndParse(control, accBits, chainIdBits));
+        return (nonce, epoch, attesterId, revealNonce, chainId);
```
vimwitch, line 83: I think if you use named variable in the function definition, like `returns (uint varA, uint varB)`, you don't need to explicitly return the value. So here this line can be removed.

packages/contracts/contracts/Unirep.sol
```diff
@@ -434,22 +443,52 @@ contract Unirep is IUnirep, VerifySignature {
 
     function decodeSignupControl(
         uint256 control
-    ) public pure returns (uint160 attesterId, uint48 epoch) {
+    ) public pure returns (uint160 attesterId, uint48 epoch, uint48 chainId) {
+        chainId = uint48((control >> (208)) & ((1 << 36) - 1));
         epoch = uint48((control >> 160) & ((1 << 48) - 1));
         attesterId = uint160(control & ((1 << 160) - 1));
-        return (attesterId, epoch);
+        return (attesterId, epoch, chainId);
```
vimwitch, line 450: Here too

packages/core/src/UserState.ts
```diff
@@ -90,10 +95,13 @@ export default class UserState {
         }
         this._id = id
         this._prover = prover
+        this._chainId = -1 // need to be setup in async function
     }
 
     async start() {
         await this.sync.start()
+        const { chainId } = await this.sync.provider.getNetwork()
```
vimwitch, line 103: Maybe use `this._checkChainId`?

vimwitch (commented):
I think this is good. The tuple return values in Solidity can cause a bug if the order of the variables is not the same as they're defined so it's good to remove the explicit return.

packages/core/src/UserState.ts
vivianjeng: Thanks 🙏🏻
[3d0902b](https://github.com/Unirep/Unirep/pull/547/commits/3d0902b9534b5a456177cf90c394117957f9a23d)

packages/contracts/contracts/verifierHelpers/BaseVerifierHelper.sol
vivianjeng: Good suggestion 🙏🏻
[d10d36b](https://github.com/Unirep/Unirep/pull/547/commits/d10d36b975a3c551d8fd9a9bdcb2d49eec190379)

vimwitch (approved):

PR #567, opened 2023-10-26
feat: add genesis block as configuration in synchronizer

ryanycw:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

In the current Synchronizer design, `_findStartBlock` will always start querying from block `0x0`.

## What is the new behavior?

In this pull request, we add a configuration variable `genesisBlock`. There are two benefits for the application that inherits this change. One is that by specifying the genesisBlock, they can shorten the setup time since the Synchronizer can now query from a closer block instead of block 0. The second benefit is that DevOps tools like Tenderly only support forking from the latest 1000 blocks at this moment. With this change, we can integrate Tenderly DevNet into the CICD deployment.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/Synchronizer.ts
+++ b/packages/core/src/Synchronizer.ts
@@ -131,7 +131,7 @@ export class Synchronizer extends EventEmitter {
      * The genesis block of the provider environment
      * @dev Allow passing the genesis block as the starting block for querying events
      */
-    public genesisBlock: number = 0
+    private _genesisBlock: number = 0
 
     /**
      * @private
@@ -193,7 +193,7 @@ export class Synchronizer extends EventEmitter {
             provider
         )
         this._provider = provider
-        this.genesisBlock = genesisBlock ?? 0
+        this._genesisBlock = genesisBlock ?? 0
         this._settings = {
             stateTreeDepth: 0,
             epochTreeDepth: 0,
@@ -310,6 +310,13 @@ export class Synchronizer extends EventEmitter {
      */
     get settings() {
         return this._settings
+    }
+
+    /**
+     * Read the genesis block of the provider environment.
+     */
+    get genesisBlock() {
+        return this._genesisBlock
     }
 
     /**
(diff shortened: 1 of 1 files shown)


packages/core/src/Synchronizer.ts
```diff
@@ -127,6 +127,11 @@ export class Synchronizer extends EventEmitter {
      * How many blocks the synchronizer will query on each poll. Default: `100000`
      */
     public blockRate: number = 10000
+    /**
+     * The genesis block of the provider environment
+     * @dev Allow passing the genesis block as the starting block for querying events
+     */
+    public genesisBlock: number = 0
```
vivianjeng, line 134: It can be `private _genesisBlock` and we use a getter to access the data
e.g.
https://github.com/Unirep/Unirep/blob/2e9e45969cd276695ea6802876b31706f2d015e6/packages/core/src/Synchronizer.ts#L283

So we won't change this value unexpectedly

vivianjeng (commented):
Thank you for your contribution 🙏🏻

packages/core/src/Synchronizer.ts
ryanycw: Thank you for the feedback. This, indeed is a better style. I've fixed it!
Looking forward to any other feedback and suggestions.

packages/core/src/Synchronizer.ts
```diff
@@ -127,6 +127,11 @@ export class Synchronizer extends EventEmitter {
      * How many blocks the synchronizer will query on each poll. Default: `100000`
      */
     public blockRate: number = 10000
+    /**
+     * The genesis block of the provider environment
+     * @dev Allow passing the genesis block as the starting block for querying events
```
vivianjeng, line 132: Actually this is not like solidity to write `@dev`
It is in typedoc format, you can check the reserved word and usage here: https://typedoc.org/guides/overview/

I prefer just written for developers
so it will be like
```ts
/**
 * Allow passing the genesis block as the starting block for querying events
 */
```

vivianjeng (commented):
sorry just found another issue 🙏🏻
Please check

packages/core/src/Synchronizer.ts
ryanycw: Thank you for pointing out this issue. It is great to learn this.

vivianjeng (approved):
Thank you for your contribution 🥳

